PR #1, opened 2017-09-21
README: document build and install processes

gsanchietti:


README.rst
```diff
@@ -0,0 +1,99 @@
...
+================
+
+Enter API directory and build using grunt: ::
+
+  cd api/system/ && grunt build
+
+
+Enter UI directory and build using grunt: ::
+
+  cd ui/ && grunt build && bower install
```
edospadoni: I think is enough `cd ui/ && grunt buil` instead of `cd ui/ && grunt build && bower install`, because the `bower_components` directory is already filled in the previous step.

gsanchietti:
Fixed.

PR #32, opened 2018-12-30
substitute globally

mikeBOLD:
Dig fails and the hint is always shown if there are more than two DNS servers in the configuration database because only the first ',' was substituted.


gsanchietti:
Really good catch, thank you!

PR #33, opened 2019-01-03
Adjust correct name for legacy in application menu

stephdl:
the rpm is not listed in the application menu, we need to set the correct rpm name.


stephdl:
@gsanchietti @edospadoni do I need to create an issue to launch a QA , or the fix is enough trivial to avoid a QA ?

PR #34, opened 2019-01-05
Implement perl function write_json

stephdl:
We need to manipulate json data, this function will help to do it

### DIFF
--- a/api/lib/auth_functions.pl
+++ b/api/lib/auth_functions.pl
@@ -57,11 +57,10 @@ sub write_json {
     my $umask = shift || 022;
 
     umask $umask;
-    open my $fh, ">", $filePath || return 0;
+    open my $fh, ">", $filePath;
     print $fh encode_json($data);
     close $fh;
-    return 1;
-}
+    }
 
 sub whoami
 {
--- a/api/lib/auth_functions.pl
+++ b/api/lib/auth_functions.pl
@@ -50,16 +50,17 @@ sub write_json {
     my $filePath = shift || die 'No file path specified';
     my $dirname = dirname($filePath);
     if (! -d $dirname) {
-        die 'No directory to save the json file';
+        warn 'No directory to save the json file';
     }
     
-    my $data = shift || die 'No data to encode to json';
+    my $data = shift || warn 'No data to encode to json';
     my $umask = shift || 022;
 
     umask $umask;
-    open my $fh, ">", $filePath;
+    open my $fh, ">", $filePath || return 0;
     print $fh encode_json($data);
     close $fh;
+    return 1;
 }
 
 sub whoami
--- a/api/lib/auth_functions.pl
+++ b/api/lib/auth_functions.pl
@@ -24,6 +24,7 @@ use strict;
 use warnings;
 use JSON;
 use POSIX qw(getgroups);
+use File::Basename;
 
 sub read_json
 {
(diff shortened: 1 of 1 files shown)


edospadoni:
Where you need to use this feature?

stephdl:
Here, to write and encode the `/etc/nethserver/cockpit/authorization/roles.json`
- update role
https://gist.github.com/stephdl/4d85360c222fec28bbdda1d4bd128027
- delete role
https://gist.github.com/stephdl/8871a162f2cae75c56f2e708b91b1023

The idea is to made some wrappers and ease the development, for now I use myself the code in two places, hence the idea to make a function.


cc @gsanchietti

edospadoni:
Maybe it's better to move the `write_json` function in `/usr/libexec/nethserver/api/lib/helper_functions.pl` instead of `/usr/libexec/nethserver/api/lib/auth_functions.pl`, and also the `read_json`.

We need to check all the occurrences in API scripts.

api/lib/auth_functions.pl
```diff
@@ -41,6 +42,26 @@ sub read_json
     return decode_json($json);
 }
 
+# encode json data and write to file ($umask optional, default 022)
+# do: write_json ($filePath,$data,$umask)
+
+sub write_json {
+
+    my $filePath = shift || die 'No file path specified';
```
gsanchietti: Functions inside a library should avoid ending the whole script.
Please gracefully handle any error.

api/lib/auth_functions.pl
```diff
@@ -41,6 +42,26 @@ sub read_json
...
+    my $dirname = dirname($filePath);
+    if (! -d $dirname) {
+        die 'No directory to save the json file';
+    }
+    
+    my $data = shift || die 'No data to encode to json';
+    my $umask = shift || 022;
+
+    umask $umask;
+    open my $fh, ">", $filePath;
```
gsanchietti: I'd change to something like:
```
 open my $fh, ">", $filePath || return 0;
 ...
 return 1
```

stephdl: why do you want to return 1 on success and 0 on error :-?

gsanchietti: Yes, it's usually how functions are written: return true on success, false otherwise.
You could also return an object on success and undef on failure.

The semantic of values returned from functions is usually the opposite of shell exit code ;)

stephdl:
```
# encode json data and write to file ($umask optional)
# write_json ($file,$data,$umask)
sub write_json {

      my $file = shift || warn 'No specified file to write';
      my $dirname = dirname($file);
      if (! -d $dirname) {
          warn 'No directory to save the json file'
      }
      my $data = shift || warn 'No data to encode to json';
      my $umask = shift || 022;

      umask $umask;
      open my $fh, ">", $file || return 0;
      print $fh encode_json($data);
      close $fh;
      return 1;
}

do you agree with this @gsanchietti , do we need to move it to `helper_functions.pl`
```

gsanchietti:
Yes, it's fine.
For now, leave the function inside auth_functions.pl

stephdl:
this PR is obsoleted by https://github.com/NethServer/nethserver-cockpit/pull/38/commits/8357bdfeebd19bff5d95c1edc38009186e4d21d6#diff-3ae22c3f8e15576ccee99d6425499dc3R45

PR #37, opened 2019-01-09
Update some messages

filippocarletti:


ui/src/i18n/locale-en.json
```diff
@@ -858,7 +858,6 @@
       "valid_platform,ldap-credentials,ldaptestbind,32": "Base DN was not found or insufficient rights",
       "valid_platform,ldap-credentials,ldaptestbind,34": "Invalid DN syntax",
       "active_directory_probe_failed": "Active directory probe failed",
-      "join_failed": "join_failed",
```
stephdl: Fun that it is not used, bug or future feature ?

PR #38, opened 2019-01-09
dev on delegation of authority (WIP)

stephdl:
to be continued, template side is not  done, I need some review before


gsanchietti:
Is this an implementation of delegation panel? In this case, why it is coded inside the main group page?

I thought about a new page, but integrating it inside the group page could be a good solution.
Any though on that @edospadoni , @DavidePrincipi @filippocarletti ?

(P.S. I just removed the reviewer: this is much more a WIP than something ready to be reviewed)

stephdl:
API, cockpit vue.js, esmith templates done, maybe time to ask a your impressions and get more directives where to go

@edospadoni @gsanchietti

root/etc/e-smith/templates/etc/sudoers.d/30nethserver_cockpit_roles/10base
```diff
@@ -0,0 +1,99 @@
+{
+
+    use strict;
+    use warnings;
+    use JSON;
+    #use esmith::ConfigDB;
+
+
+    sub read_json
```
edospadoni: remove read_json and use `require "/usr/libexec/nethserver/api/lib/auth_functions.pl";`

api/system-roles/read
```diff
@@ -0,0 +1,61 @@
...
+
+use warnings;
+use strict;
+use JSON;
+
+my $input = readInput();
+my $role = $input->{'role'} || '';
+my $cmd = $input->{'action'} || '';
+
+if ($cmd eq 'listApplications') {
```
gsanchietti: Do not use camel case (just like other actions).

A good choice could be "applications".

api/system-roles/read
```diff
@@ -0,0 +1,61 @@
...
+    foreach my $a (@apps) {
+        push($ret->{"applications"}, $a);
+    }
+
+    print encode_json($ret);
+    exit(0);
+
+}
+
+my $data = read_json("/etc/nethserver/cockpit/authorization/roles.json");
```
gsanchietti: Add another action here, so the client must always specify it.
Something like "roles".

api/system-roles/read
```diff
@@ -0,0 +1,61 @@
...
+
+if(exists($data->{$role})) {
+    print encode_json ($data->{$role});
+}
```
gsanchietti: Change the indentation:
```
if (..) {
    ....
} else {
}
```

api/system-roles/update
```diff
@@ -0,0 +1,48 @@
...
+    $data->{$role}->{'system'} = $input->{'system'};
+    $data->{$role}->{'applications'} = $input->{'applications'};
+    write_json ($jsonFile,$data);
+    my $out = `/sbin/e-smith/expand-template /etc/sudoers.d/30nethserver_cockpit_roles`;
+    if ($? > 0) {
+        error("EventFailed", "");
+    } else {
+        success();
+    }
+} else {
```
gsanchietti: Indentation, same as above.

root/etc/e-smith/templates/etc/sudoers.d/30nethserver_cockpit_roles/10base
```diff
@@ -0,0 +1,99 @@
...
+
+        return decode_json($json);
+    }
+
+
+
+    my $jsonFile = '/etc/nethserver/cockpit/authorization/roles.json';
+    my $data = read_json ($jsonFile); 
+
+    foreach my $key (keys %{$data}) {
```
gsanchietti: I would love to see a map instead a set of nested if.

A map where you have the view as the key, and the controllers as values.

Something like:
```perl
%controllers = (
  "users" => ("system-roles", "system-password-policy"),
   ...
);
```

root/etc/e-smith/templates/etc/sudoers.d/30nethserver_cockpit_roles/10base
```diff
@@ -0,0 +1,99 @@
...
+                $role = 'system-openssh';
+            }
+            elsif ($role eq 'system-applications') {
+                $role = 'system-apps';
+            }
+            elsif ($role eq 'system-software-center') {
+                $role = 'system-packages';
+            }
+
+            $role = '*' if ($key eq 'admins');
```
gsanchietti: Instead of "admins", we should use the value inside the `admins` key.
The default is `domain admins`, we should than change the default `roles.json` (I guess @edospadoni already did it).

stephdl: in fact this part should be removed @gsanchietti , For the administrator group we should use the same way than for root and return the full list of applications, rather than a list of role. This is the rebrand I would like to do after this PR

edospadoni: yes, job done

gsanchietti:
# Main menu

There is no filter on the main left menu.
A user which is not granted access to an item (eg. Applications) should see the link.

Also the left menu items (Applications, Subscription. etc) should be added inside the `roles.json`.

## Possible solution 1

Create a copy of nethserver cockpit application using some links inside the user home directory, than create a new manifest.json which overrides the default one.

Example, to be executed as user `giacomo`:
```
mkdir -p .local/share/cockpit/nethserver
cd .local/share/cockpit/nethserver
ln -sf /usr/share/cockpit/nethserver/applications/
ln -sf /usr/share/cockpit/nethserver/categories/
ln -sf /usr/share/cockpit/nethserver/index.html
ln -sf /usr/share/cockpit/nethserver/libs/
ln -sf /usr/share/cockpit/nethserver/static/
```

Content of .local/share/cockpit/nethserver/manifest.json: 
```
{
  "version": 0,
  "name": "nethserver",

  "menu": {
    "index": {
      "label": "System",
      "order": 10
    },
    "#/applications": {
      "label": "Applications",
      "order": 20
    },
    "///#/terminal": {
      "label": "Terminal",
      "order": 50
    }
  },

  "content-security-policy": "default-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' data:"
}
```
Above commands should be executed when creating the delegation policy for each user inside the group.

## Possible solution 2

Implement a UI trick to hide some menu items upon login.

# Bookmarks

Users should not be able to see the bookmarks on the left corner.
Remove `tools` section under manifest.json.

# User password change

Current implementation of non-root user password change is completely broken because relays on `passwd` command which can't be used with current sssd configuration.

This part should be completely rewritten.

## Possible solution 1

Implement password change using LDAP client, such solution should work with local and remote account providers but requires access to system configuration.
Account provider parameters, such as URI, must be exported in a configuration file and accessible to any user.


## Possible solution 2

Create a wrapper for `password-change` policy to be executed using sudo. This an high risk task because an implementation error could cause a dangerous local privilege escalation.

# Conventions

The `/etc/sudoers.d/30nethserver_cockpit_roles` template should be renamed to `/etc/sudoers.d/30_nethserver_cockpit_roles`.

gsanchietti:
@DavidePrincipi  is working on password user change which totally broken.

stephdl:
obsoleted by https://github.com/NethServer/nethserver-cockpit/pull/50

PR #51, opened 2019-02-15
Allow simple user to login in cockpit

stephdl:


root/etc/e-smith/templates/etc/shells/10base
```diff
@@ -0,0 +1,6 @@
+/bin/bash
```
gsanchietti: I'd like to move lines from 1 to 6 in a fragment named something like "10base"  or "10default".

And the line 6 to a fragment named something like "20nethserver"

ui/package-lock.json
```diff
@@ -4944,7 +4944,7 @@
     },
     "http-proxy-middleware": {
       "version": "0.18.0",
-      "resolved": "https://registry.npmjs.org/http-proxy-middleware/-/http-proxy-middleware-0.18.0.tgz",
+      "resolved": "http://registry.npmjs.org/http-proxy-middleware/-/http-proxy-middleware-0.18.0.tgz",
```
gsanchietti: Please revert this change.

PR #53, opened 2019-02-20
fixes to make "npm run dev" pass

dasois:
The build failed on my machine until i fixed those issues.
Travis does not build webpack, right? Maybe we should add that....

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -89,12 +89,6 @@
       "integrity": "sha1-l6ERlkmyEa0zaR2fn0hqjsn74KM=",
       "dev": true
     },
-    "ansi-colors": {
-      "version": "3.2.3",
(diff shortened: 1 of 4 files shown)


edospadoni:
> Travis does not build webpack, right? Maybe we should add that....

Travis **builds** webpack using the `prep-sources` script.

It is invoked during `install` step inside travis:
https://github.com/NethServer/nethserver-cockpit/blob/master/.travis.yml#L39

It seems that the build is not broken, the .rpm is builded and you can see it above in the messages.

Ok for other fixes :+1: , thank you.

dasois:
Thanks for your feedback, of course I should have made a separate branch for that. 
I will take back that PR, since Travis does build npm run dev and thus there seemed to be something wrong with my machine. Investigating...

PR #55, opened 2019-02-21
Fix: Dev Environment Packages

dasois:
In order to get "npm run dev" to work, the packages needed to be changed. Webpack-dev-server was downgraded according to https://stackoverflow.com/questions/53690282/typeerror-cannot-destructure-property-compile-of-undefined-or-null

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -89,12 +89,6 @@
       "integrity": "sha1-l6ERlkmyEa0zaR2fn0hqjsn74KM=",
       "dev": true
     },
-    "ansi-colors": {
-      "version": "3.2.3",
-      "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-3.2.3.tgz",
-      "integrity": "sha512-LEHHyuhlPY3TmuUYMh2oz89lTShfvgbmzaBcxve9t/9Wuy7Dwf4yoAKcND7KFT1HAQfqZ12qtc+DUrBMeKF9nw==",
-      "dev": true
-    },
     "ansi-escapes": {
       "version": "1.4.0",
       "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-1.4.0.tgz",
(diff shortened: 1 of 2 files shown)


edospadoni:
I'm not sure that this fix is necessary, what kind of error do you receive in you local dev environment?

dasois:
Exactly the one from the referenced StackOverflow site:
`TypeError: Cannot destructure property `compile` of 'undefined' or 'null'`

edospadoni:
In my machine with:
```
node -v
v10.15.1
```

I have no problems building the repo

dasois:
I have `v8.15.0`.
`npm audit` doesn't report an error on the master branch. 
On this change-branch however i get:
                                                                                
```
                       === npm audit security report ===                        
                                                                                
# Run  npm install --save-dev webpack-dev-server@3.2.0  to resolve 2 vulnerabilities
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ braces                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ webpack-dev-server [dev]                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ webpack-dev-server > http-proxy-middleware > micromatch >    │
│               │ braces                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/786                             │
└───────────────┴──────────────────────────────────────────────────────────────┘


┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Missing Origin Validation                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ webpack-dev-server                                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ webpack-dev-server [dev]                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ webpack-dev-server                                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/725                             │
└───────────────┴──────────────────────────────────────────────────────────────┘


found 2 vulnerabilities (1 low, 1 high) in 10406 scanned packages
  run `npm audit fix` to fix 2 of them.

```
These vulnerabilities can be ignored, since they only affect the dev-tools..

edospadoni:
Ok, the problem is not visible during the build process because affect the `webpack-dev-server` module that is not used in `npm run build`. With you fix we introduce an High vulnerability in a module that at the moment, also for the development is not used, because you have to **build** your project and sync it to the machine. 

We can archive this pr at the moment.

dasois:
Tried with node 10 now, works. 
We should update the dev-documentation..

edospadoni:
Yes, good point ;) thank you.

dasois:
damn, i reencountered this problem even with node 10 now...
what is the preferred way to build a dev-build, that is debuggable?

edospadoni:
At the moment the fastest way is build your package in your local machine and sync it to nethserver virtual machine. I don't find a smarter solution :(.

Personally I'm using a script like this
```bash
#!/bin/bash
while true
  do
    inotifywait -r -e modify,attrib,close_write,move,create,delete . && npm run build && rsync ...
done
```
that build and sync the `dist/` directory every time you save files inside the project, but is not the best solution

dasois:
Thanks for pointing this out! :)
I meant: How can i build a non-production build with non-minified sources so I'm able to debug in the webbrowser?

edospadoni:
Sorry, I did not understand :)

You can debug in the webbrowser expanding the file inside the developer tools, under `Source`. YOu can select the correct frame and click on `webpack://`. Webpack create a map to handle the minified files. 

But maybe there is an option to bypass the minification, have you try to search inside the official doc?

PR #56, opened 2019-02-21
Fix: Close b-tag

dasois:
<b>-tag was not closed, slash was missing.


dasois:
Only `npm run dev` discovered this. That's why I thought at first, that the vue-build doesn't get executed by Travis. Maybe some investigation is needed, why the production build ignores this error...

PR #57, opened 2019-02-21
Fix: v-model conflict with :value on input

dasois:
`:value` was not necessary, because value already bound by v-model. This broke dev-build.


dasois:
Only `npm run dev` discovered this. That's why I thought at first, that the vue-build doesn't get executed by Travis. Maybe some investigation is needed, why the production build ignores this error...

PR #58, opened 2019-02-21
Fix: Calculation of Software Updates Count

dasois:
The counter was not actually counting the number of update-packages, but the number of update categories (nethserver, other)...


gsanchietti:
Excellent, thank you!

PR #61, opened 2019-02-22
Do not allocate a pty in nethserver.exec()

DavidePrincipi:
A terminal allocation is usually not required to send data to a process.
Only few programs want to read from a terminal device (i.e. password
change). More information https://linux.die.net/man/4/pts

### DIFF
--- a/root/usr/share/cockpit/nethserver/libs/nethserver.js
+++ b/root/usr/share/cockpit/nethserver/libs/nethserver.js
@@ -21,10 +21,7 @@ nethserver = {
             args[0] = "/usr/libexec/nethserver/api/" + api
         }
 
-        var process = cockpit.spawn(args, input ? {} : {
-            pty: true,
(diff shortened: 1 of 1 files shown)


DavidePrincipi:
@edospadoni 

- password change (both root-self and other user's one) works
- software center works (with #65)

gsanchietti:
Please try also try the execute methods, like the "follow log" and few others around.

DavidePrincipi:
Logs (mattermost, system) work. Also accounts provider installation.

edospadoni:
Try also the `format USB` action inside the backup. When you choose a USB location as destination

DavidePrincipi:
Format USB works too!

gsanchietti:
So it should be safe enough!

edospadoni:
Ok for me, we can remove the PTY

PR #62, opened 2019-02-22
Add Bootstrap Switch and TouchSpin widgets

DavidePrincipi:
The patternfly CSS already defines the styles for both Bootstrap Switch and TouchSpin widgets. Only the JS code is missing.

Why not adding the widgets code in a common library?

This is required by https://github.com/NethServer/nethserver-ejabberd/pull/20 but is likely to be used by other packages too.


DavidePrincipi:
This is required by https://github.com/NethServer/dev/issues/5719

PR #80, opened 2019-03-20
Return the LANG env variable value

DavidePrincipi:


ui/src/main.js
```diff
@@ -38,20 +38,22 @@ Vue.mixin(NethServer)
...
-  locale: langConf.locale,
-  messages: langConf.messages
-});
-var moment = require("moment");
-moment.locale(langConf.locale);
+const i18n = new VueI18n();
 
 // init Vue app
 var ns = new Vue({
-  el: "#app",
```
DavidePrincipi: the `el` attribute actually runs `mount()` immediately: must be removed because language.json is still not loaded

gsanchietti:
Merge after security fix has been disclosed.

PR #84, opened 2019-04-04
api system-logs: add 'filter' field

gsanchietti:
Allow log filter, useful for the mail application.


gsanchietti:
You're already root, I wouldn't mind much about shell injection.

By the way, I added shell quote check. I'm avoiding the array because I need to pipe 2 commands.

DavidePrincipi:
> You're already root

I was thinking about non-root domain admins...

> I need to pipe 2 commands

lazy guy :smiley_cat:

PR #90, opened 2019-04-30
Override JSON manifest at runtime

DavidePrincipi:
The `infoapi` attribute causes the given API script to be invoked. Its output
is merged with the JSON manifest contents, to allow runtime overrides.

The API script receives a JSON object as input. The input object is obtained
by the merge of:

1. system-apps/read input object, where the action value is forcibly set
   to "app-info"

2. the `infoapi.input` object

### DIFF
--- a/api/system-apps/read
+++ b/api/system-apps/read
@@ -46,11 +46,45 @@ sub read_json
     return decode_json($json);
 }
 
+sub invoke_info_api
+{
+    use IPC::Open2;
+    my $helper_path = shift;
+    my $helper_input = shift;
+    my($chld_out, $chld_in, $pid);
+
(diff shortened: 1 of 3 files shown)


api/system-apps/read
```diff
@@ -46,11 +46,45 @@ sub read_json
...
 sub app_info
 {
     my $app = shift;
     my $file = "$path$app.json";
     my $data = read_json($file);
+    my $extended_data = {};
+
+    # If available, invoke the INFO_API provider
+    # to extend the manifest data dynamically
+    if($data->{'INFO_API'}) {
```
gsanchietti: I'd change the key to `infoapi` or `info_api`. Everything else is in lower case, and I'd like to respect the conventions.

PR #91, opened 2019-05-02
Allow subfolder delegation in API

stephdl:
With the new way to organize the API, we can have some some API subfolder, for example

nethserver-mail/domain/update

The commit is there to allow all subfolders instead to allow only one level of folder (nethserver-mattermost/update for example)


stephdl:
this will need  to give more granular sudo permission from the developer itself, we need to find the good implementation, lets close it and find the good one

PR #96, opened 2019-05-09
api system-users: display lastLogon

gsanchietti:
- Display lastLogon for users if account provider is Active Directory

NethServer/dev#5754


gsanchietti:
This should not implemented since the value seems not reliable.

PR #104, opened 2019-06-19
Added support for pinned applications

faffolao:
See also NethServer/nethserver-httpd#34

### DIFF
--- a/docs/docs/api/nethserver-suricata/bypass.md
+++ b/docs/docs/api/nethserver-suricata/bypass.md
@@ -1,0 +1,214 @@
+# bypass
+
+Read and set bypass for suricata.
+
+## read
+
+The read API takes an `action` field.
+
+Valid actions are:
+
+- `bypass-list`
+- `object-list`
+
+#### bypass-list
+
+List all configured bypasses.
+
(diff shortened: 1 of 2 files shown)


ui/public/i18n/language.json
```diff
@@ -45,6 +45,8 @@
   "hints_suggested": "Please, review the following settings",
   "add_shortcut": "Add shortcut",
   "remove_shortcut": "Remove shortcut",
+  "add_pin": "Add pin",
```
gsanchietti: I'd change the label to:
```
Add to home page
```

ui/public/i18n/language.json
```diff
@@ -45,6 +45,8 @@
...
+  "remove_pin": "Remove pin",
```
gsanchietti: I'd change the label to:
```
Remove from home page
```

ui/public/i18n/language.json
faffolao: updated

ui/public/i18n/language.json
faffolao: updated

PR #111, opened 2019-08-08
Provider change and password policy only for root

stephdl:
only root can change provider settings and password policy inside the user-group UI


gsanchietti:
Please add some doc about it before merging.

gsanchietti:
Let's merge as is, we will refactor the doc (https://nethserver.github.io/nethserver-cockpit/authorizations/) when all the work on the sudo part is done.

PR #112, opened 2019-08-09
Display MAC address for VLANs in Network page

andre8244:
MAC address for virtual interfaces is now retrieved and displayed in Cockpit Network page.


cotosso:
tested: it works as expected

PR #117, opened 2019-08-19
Sudoers based authorizations

stephdl:
https://github.com/NethServer/dev/issues/5805

Note: use MERGE commit - do not squash


DavidePrincipi:
In nethserver-testing:

- nethserver-cockpit-0.11.0-1.45.gb30fb18.ns7.noarch.rpm
- nethserver-cockpit-lib-0.11.0-1.45.gb30fb18.ns7.noarch.rpm

DavidePrincipi:
In nethserver-testing:

- nethserver-cockpit-0.11.0-1.54.g804ec75.ns7.noarch.rpm
- nethserver-cockpit-lib-0.11.0-1.54.g804ec75.ns7.noarch.rpm

root/etc/e-smith/templates/etc/sudoers.d/55_nsapi_perms/90admins
```diff
@@ -0,0 +1,16 @@
+#
+# 90admins
+#
+{
+    use NethServer::ApiTools;
+
+    my $adminGroup = ($admins{'group'} || 'domain admins') =~ s/ /\\ /r;
+    my $adminUser = ($admins{'user'} || 'admin') =~ s/ /\\ /r;
+
+    my $commands = join ', ', sort map { 'NSAPI_' . uc($_ =~ tr/-/_/r) } 'admins', 'public', NethServer::ApiTools::list_applications();
```
DavidePrincipi: Let's make `admin` `domain admins` a root-equivalent, like the wheel group!

```suggestion
    my $commands = 'ALL';
```

stephdl: indeed, after testing : `%domain\ admins ALL=NOPASSWD: ALL`
pro :
- we do not need to expand the template `/etc/sudoers.d/55_nsapi_perms` after each rpm installation
- admins are full granted to log viewer, storage panel and probably more administrative panel, this could be also a weak (notably for the storage panel)
- we still have the same problem for simple user to be granted to the cockpit log viewer

gsanchietti:
> In nethserver-testing:
> 
> * nethserver-cockpit-0.11.0-1.54.g804ec75.ns7.noarch.rpm
> * nethserver-cockpit-lib-0.11.0-1.54.g804ec75.ns7.noarch.rpm

The package has been removed from testing to avoid conflicts with nethserver-fail2ban: nethserver/dev#5813

root/etc/e-smith/templates/etc/sudoers.d/55_nsapi_perms/90admins
DavidePrincipi: Showstopper: the default wheel perms are

    %wheel	ALL=(ALL)	ALL

That means the members of the wheel group are asked for password to run a privileged command. I think this is an important security measure, so I'd like to remove the `NOPASSWD` flag. However this breaks the current `sudo` invocation style:

    Sep 18 10:44:56 vm5.dpnet.nethesis.it cockpit-bridge[32291]: sudo: no tty present and no askpass program 

To integrate it correctly with cockpit the "askpass" program must be properly working. I think we need to fix they way `sudo` is invoked. Cockpit should be already do it, but I prefer to avoid fixing our code by now.

Let's do it at a later step.

DavidePrincipi:
Note: use MERGE commit - do not squash

PR #122, opened 2019-08-30
AD DNS server validation

andre8244:
The validation script of Username and Password expects an AD DNS Server, but in some scenarios the user doesn't fill it.
In these cases the IP address of the `Domain name` is used as `AD DNS Server`.

https://github.com/NethServer/dev/issues/5817

### DIFF
--- a/api/system-accounts-provider/validate
+++ b/api/system-accounts-provider/validate
@@ -98,6 +98,9 @@ function validate_remotead($data) {
 
 
 function validate_adcredentials($data) {
+    if (!array_key_exists('AdDns', $data)) {
+        $data['AdDns'] = gethostbyname($data['AdRealm']);
+    }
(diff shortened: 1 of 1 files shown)


api/system-accounts-provider/validate
```diff
@@ -98,6 +98,9 @@ function validate_remotead($data) {
 
 
 function validate_adcredentials($data) {
+    if (!array_key_exists('AdDns', $data)) {
```
gsanchietti: If `AdDns` is not set, use the upstream DNS.

Something like:
```
$db = new EsmithDatabase('configuration');
$ns = explode(",",$db->getProp('dns', 'NameServers'));
if (isset($ns[0]) && $ns[0]) {
    $addns =$ns[0];
} else {
$addns = "127.0.0.1"
}

```

PR #125, opened 2019-09-07
Scan the hosts on a network interface

stephdl:
A diagtools to scan the hosts on a NIC, and make a reservation with the mac address.

If the mac address is already set in the host esmith database for another IP address, the button to add the IP reservation is disabled

https://github.com/NethServer/dev/issues/5830

### DIFF
--- a/api/system-dhcp/read
+++ b/api/system-dhcp/read
@@ -23,6 +23,9 @@
 
 require_once("/usr/libexec/nethserver/api/lib/dhcp_functions.php");
 
+$data = readInput();
+$action = $data['action'];
+
 function not_expired($r)
 {
     if (time() > $r['expire']) {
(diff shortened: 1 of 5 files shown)


api/system-dhcp/scanNetwork
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/perl
+
+#
+# Copyright (C) 2018 Nethesis S.r.l.
```
gsanchietti: Fix the year.

Also the whole implementation should go under the the `read` API with a new `scan` action (it seems the whole API should be refactored for handling multiple actions).
If you prefer to have the code in another script, invoke the helper from within read API.

(Please also note that this script is not executable inside the RPM, I had to manually fix it) 

See https://nethserver.github.io/nethserver-cockpit/api_guidelines/

api/system-dhcp/validate
```diff
@@ -51,6 +51,13 @@ if ($action == 'create-reservation') {
         $v->addValidationError('IpAddress','dhcp_reservation_system_validator');
     }
 
+    # valid if the mac address is not already reserved
```
gsanchietti: This new code shouldn't be needed since the same validation rule is already present inside the 'dhcp-reservation' system validator called above.

ui/src/components/system/DHCP.vue
```diff
@@ -41,6 +48,7 @@
           v-if="i.props.DhcpRangeStart && i.props.DhcpRangeStart.length > 0 && i.props.DhcpRangeEnd && i.props.DhcpRangeEnd.length > 0"
           class="gray margin-left-md"
         >({{i.props.DhcpRangeStart}} - {{i.props.DhcpRangeEnd}})</span>
+
```
gsanchietti: Remove this new line just for a cleaner diff

ui/public/i18n/language.json
```diff
@@ -393,7 +393,9 @@
     "range_create_ok": "Range created successfully!",
     "range_create_error": "Range not created.",
     "range_update_ok": "Range updated successfully!",
-    "range_update_error": "Range not updated."
+    "range_update_error": "Range not updated.",
+    "scan_network": "Scan hosts",
```
gsanchietti: I'd prefer "Scan network" as translation, but it's just a personal taste :)

api/system-dhcp/read
```diff
@@ -36,54 +39,82 @@ $ranges = array();
...
+    $nic = $data['nic'];
+    $hosts_db = new  EsmithDatabase('hosts');
+    $macs = array();
+
+    foreach ($hosts_db->getAll("local") as $key => $props) {
+        $mac = $hosts_db->getProp($key, 'MacAddress');
+        $macs[] = array_push ($macs,$mac);
+    }
+
+    exec("arp-scan -I $nic -lx", $output, $ret);
```
gsanchietti: It's better to protect user input with [escapeshellarg](https://www.php.net/manual/en/function.escapeshellarg.php).

PR #126, opened 2019-09-10
Search in mac array with in_array

stephdl:
https://github.com/NethServer/dev/issues/5830

Find in the mac array if the mac address is not already reserved with in_array, the IP reservation button is always false (must be true if the mac address is saved in hosts esmith database)


stephdl:
@giacomo, please merge to fix the IP reservation button if you agree

stephdl:
obsoleted by https://github.com/NethServer/nethserver-cockpit/pull/128

Set closed

PR #131, opened 2019-09-21
Replace nic vendor name by arp hostname

stephdl:
Find the hostname with arp and if we know it, replace the nic vendor name of arp-scan by the arp hostname


cotosso:
I tried the update and it works fine.
I note only 2 small things (see also attachment).

1) I would put the name obtained by arp instead of the host name (dx column) which is still empty because it is not defined in the db, so we will see both the vendor and the name arp.

2) I have a case in which the name arp is solved in some way, but on the interface it is not shown.

However, it is already OK this way too.

Sistema - nsec-primary nethesis it  4  (1)

stephdl:
case 1

just to try to explain why the domain name which came to arp -ais in description and not in the host field.

the validator of the host field expects a hostname, not a FQDN, so the correct value is host and not host.local, therefore you cannot save it as it, you have to remove all the domain part when you want to make a reservation

Moreover it is a description for me, we cannot trust the value sent by the host name itself, an argument more why I could not use it to the host field

case 2 

I need  more information, all the hostnames come from arp -a, if that command gives back `?`, it is because the hostname cannot be determined, so in that case, we give back the nic vendor name

cotosso:
thank you for the clarifications!

case 1:
I agree very much on the description field and the functionality is already fine.
The fact was that It **could** be a good thing (but not very sure about that)  to see both the name of the builder and the arp name (just in case the arp name has been modified and it's unclear, so we can look at the builder name to have some more information).
But I think you can leave it as it is.

case2:
I understood why some hosts that are solved but do not appear in the panel.
These hosts are not really present in the output of arp -a.
It is necessary that they exchange a minimum of traffic with nethserver (even a ping is sufficient) and at that point they are shown in the output of arp -a and also in the web panel.

cotosso:
At this point I think that is good to leave it that way.

PR #135, opened 2019-09-27
Delegate the AddShortcut|pin in Applications

stephdl:
Delegate the add shortcut or the pin application if at least one APP has been delegated


gsanchietti:
@DavidePrincipi suggested to limit the pin ability only to root. It could be good, but I do not have a preference.

PR #136, opened 2019-09-30
sudoers: use NSAPI_SYSTEM_APPS

gsanchietti:
Fix warning: /etc/sudoers.d/50_nsapi:33 unused Cmnd_Alias "NSAPI_SYSTEM_APPS"

Missing part from #135

### DIFF
--- a/root/etc/sudoers.d/50_nsapi
+++ b/root/etc/sudoers.d/50_nsapi
@@ -215,4 +215,5 @@ Cmnd_Alias NSAPI_ADMINS = \
     NSAPI_SYSTEM_TERMINAL, \
     NSAPI_SYSTEM_STORAGE, \
     NSAPI_SYSTEM_USERS_GROUPS, \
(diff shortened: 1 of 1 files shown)


root/etc/sudoers.d/50_nsapi
```diff
@@ -215,4 +215,5 @@ Cmnd_Alias NSAPI_ADMINS = \
     NSAPI_SYSTEM_TERMINAL, \
     NSAPI_SYSTEM_STORAGE, \
     NSAPI_SYSTEM_USERS_GROUPS, \
+    NSAPI_SYSTEM_APPS, \
     /usr/libexec/nethserver/api/system-apps/update
```
DavidePrincipi: Ok to fix the warning. Please note that the line 219 has some magic: when an user can run `system-apps/update` he's granted `isAdmin` permission at UI level too.

PR #141, opened 2019-10-02
ui & api. fix backup config restore

edospadoni:
https://github.com/NethServer/dev/issues/5853

### DIFF
--- a/api/system-backup/execute
+++ b/api/system-backup/execute
@@ -35,10 +35,17 @@ function _restore
 {
     id=$1
     # transform remap map to a comma-separated list
-    remap=$(echo $2 | jq -r ". | to_entries[] | [.key, .value] | join(\",\")")
+    remap=$(echo $2 | jq -r ". | to_entries[] | [.key, .value] | join(\",\")" | tr '\n' , | sed 's/,$//')
     if [ ! -z $remap ]; then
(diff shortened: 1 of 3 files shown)


api/system-backup/execute
```diff
@@ -47,7 +54,8 @@ function _restore
     echo '{"progress":"0.50","time":"2.00","exit":0,"event":"restore-config","state":"done","step":1,"pid":"","action":"Pulling from history"}';
 
     echo '{"step":2,"pid":"","action":"Restoring configuration","event":"restore-config","state":"running"}'
-    /sbin/e-smith/restore-config $remap &>/dev/null
+    echo "{\"status\": \"/sbin/e-smith/restore-config $remap $reinstall\"}"
+    /sbin/e-smith/restore-config $remap $reinstall &>/dev/null
```
DavidePrincipi: Can we preserve stderr?

```suggestion
    /sbin/e-smith/restore-config $remap $reinstall >/dev/null
```

gsanchietti: If stderr shouldn't be catch, please be sure that the UI doesn't break.

DavidePrincipi: I guess stderr is drained to system journal

PR #145, opened 2019-10-08
Cancel go back to scanNetworkModal

stephdl:
When you issue a large scan of a network, you might want to come back to the scan modal if you press `cancel`.

scan then make a reservation, if you press cancel, you go back to the IP table of scan modal
scan then make a reservation, if you press save, you go back to the dhcp main menu

we could go back also to the scan modal after we have saved the reservation

what do you think ?

### DIFF
--- a/ui/src/components/system/DHCP.vue
+++ b/ui/src/components/system/DHCP.vue
@@ -229,7 +229,8 @@
 
             <div class="modal-footer">
               <div v-if="newReservation.isLoading" class="spinner spinner-sm form-spinner-loader"></div>
-              <button class="btn btn-default" type="button" data-dismiss="modal">{{$t('cancel')}}</button>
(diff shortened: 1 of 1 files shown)


filippocarletti:
We could avoid repeating "scan" which is slow and creates network traffic.
Always back tp scan modal, then.

stephdl:
I will code to go back to the scan modal when we make a reservation from the scan modal, scan is long, we can save time

stephdl:
Scan the network, press to reserve the IP, you can go back to the scanned list
Scan the network, press to reserve the IP, make a reservation you go back to the scanned list, you are able to make another reservation

gsanchietti:
Released: nethserver-cockpit-1.1.1-1.ns7.noarch.rpm

PR #157, opened 2019-10-29
ui. Reword backup-data start message

filippocarletti:
When the user clicks on the Backup "Run now" button, the messages says
that the backup as been "executed" but it has only been started and it
may run for quite a long time (user has feedback).


gsanchietti:
NethServer/dev#5876

PR #170, opened 2019-11-22
ui. change document title on route change

edospadoni:
Nethserver/dev#5946


ui/src/App.vue
```diff
@@ -321,6 +320,14 @@ export default {
 
     // get hints
     this.checkHints();
+
+    // set document title
+    var hostname = window.top.document.title.split("-")[1].trim();
```
federicoballarini: If the hostname contains a "-" it will show only the first part.
I suggest you to change this line with:

`var hostname = window.top.document.title.split(/-(.+)/)[1].trim();`

PR #172, opened 2019-11-23
ui. show installed packages and improve software center

federicoballarini:
Refer to https://github.com/NethServer/dev/issues/5950

### DIFF
--- a/api/system-packages/list-updates
+++ b/api/system-packages/list-updates
@@ -129,4 +129,3 @@ def main():
 
 if __name__ == "__main__":
     main()
-
--- a/api/system-packages/list-updates
+++ b/api/system-packages/list-updates
@@ -88,7 +88,7 @@ def list_updates(yb):
     for p in yb.doPackageLists(pkgnarrow='installed').installed:
         installed_versions[p.name] = (p.version, p.release)
 
-    updates = map(lambda i: {'name': i.name + ' ' + i.version, 'arch': i.arch, 'epoch': i.epoch, 'installed_version': installed_versions[i.name][0], 'installed_release': installed_versions[i.name][1], 'version': i.version, 'release': i.release, 'repo': i.repoid}, yb.doPackageLists(pkgnarrow='updates').updates)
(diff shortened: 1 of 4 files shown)


federicoballarini:
Next steps:
- installed packages modal
- insert language strings

gsanchietti:
This is a preview:
Screenshot from 2019-11-25 15-47-13

I think there are still some issue:
- the number on the right is not the release, but the full version
- previously, the current installed version was installed, now it's missing
- the button to open the list of installed packages works very good, but it's the in the wrong position: all existing buttons are relative to updates. I think the button for installed packages should be moved somewhere else

federicoballarini:
_> the number on the right is not the release, but the full version_
Yes, we can correct it.

_> previously, the current installed version was installed, now it's missing_
I removed it, because of the new "Installed" button, but we can add it on the right of Full Version like:(Current: 2.4.0-1.ns7)

_> the button to open the list of installed packages works very good, but it's the in the wrong position: all existing buttons are relative to updates. I think the button for installed packages should be moved somewhere else_
I can try to find another place.

Do you think it's better to revert all my commits and push only 2 new commit?

edospadoni:
Here my reviews: 
1. The console show this error:
```
vue.runtime.esm.js?2b0e:619 [Vue warn]: Error in render: "TypeError: Cannot read property 'length' of undefined"
```

You have to declare a `packages: []` on line `562`, under the `updates` object.

----

2. When opening the Updates accordion, at the bottom you should show also the correct number of base system update, like for nethserver. The label says **2** updates, but the modal shows **10** updates
mismatch

----

3. I agree with @gsanchietti, you should move the `Installed` button somewhere else, maybe in the `Applications` page, with all installed software, but I'm not sure :thinking: . Maybe some mockups should helps

----


4. I like this idea to keep the current version
> I removed it, because of the new "Installed" button, but we can add it on the right of Full Version like:(Current: 2.4.0-1.ns7)

federicoballarini:
I will close this PR and opened new one without unuseful commits.

PR #174, opened 2019-11-26
ui. improve software center view

federicoballarini:
Refer to https://github.com/NethServer/dev/issues/5959

### DIFF
--- a/api/system-packages/read
+++ b/api/system-packages/read
@@ -88,6 +88,11 @@ if($cmd eq 'list-updates') {
     }
     print $out;
 
+} elsif ($cmd eq 'list-installed') {
+
+    my $out = `/bin/rpm -qa --queryformat '%{NAME}:%{VERSION}:%{VERSION}-%{RELEASE}\n' | sort | column -t -s :`;    
+	print $out;
+
 } elsif ($cmd eq 'list-removed') {
 
     my $packages = $input->{'packages'};
--- a/api/system-packages/read
+++ b/api/system-packages/read
@@ -90,8 +90,8 @@ if($cmd eq 'list-updates') {
 
 } elsif ($cmd eq 'list-installed') {
 
-    my $out = `/bin/rpm -qa --queryformat '%{NAME}:%{VERSION}:%{VERSION}-%{RELEASE}\n' | sort | column -t -s :`;    
-	print $out;
+    my $out = `/bin/rpm -qa --queryformat '%{NAME}:%{VERSION}:%{VERSION}-%{RELEASE}\n' | sort | column -t -s :`;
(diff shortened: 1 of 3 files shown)


federicoballarini:
@gsanchietti It should be fine, but rpm fails also on my local repo without my commits. Can you check?

federicoballarini:
I think that now can be ok: @gsanchietti and @edospadoni what do you think?

- Updates count is now ok
- moved Installed Packages button into Applications section
- Installed packages details modal works fine
- Updates details updated

Schermata 2019-11-27 alle 14 24 15
Schermata 2019-11-27 alle 14 28 13

edospadoni:
Yes, better. Just one thing. I prefer to change the "Installed package" button to a simply `<a>` tag instead of a button. Like the `Details` below

federicoballarini:
In which position do you want to place the <a> tag? I like the idea of button.

edospadoni:
In the same position of button, but instead of a button change to `<a>` tag, a link that opens the modal

federicoballarini:
Like this? I moved it on the left.

Schermata 2019-11-27 alle 15 03 28

But I still prefer the button :)

edospadoni:
You're right, it's better the button. Keep it on the left position, but try to reduce the size by removing the `btn-lg` class.

federicoballarini:
This is the button without `btn-lg` class. What do you think? We can remove it or do you prefer to maintain it?

Schermata 2019-11-27 alle 15 30 58

edospadoni:
We can remove it, because the primary action must be `Install x applications`. The normal size fits the other buttons in the page.

federicoballarini:
I changed btn class and color of current full version as follow:

Schermata 2019-11-27 alle 16 22 21

edospadoni:
Nice ;)

edospadoni:
Just 2 things:
- handle long description: the text overflows on `Details` label (see image attached)
- change `Cancel` to `Close` in modals: **Installed packages** and **Changelog**

error_long

federicoballarini:
This is the interface now:

Schermata 2019-11-28 alle 00 37 20

And while updating: 

Schermata 2019-11-28 alle 00 39 34

Changed also "Cancel" with "Close".
What do you think?

federicoballarini:
This is new look:

Schermata 2019-11-28 alle 01 04 34

And while updating:

Schermata 2019-11-28 alle 01 04 43

What do you think @edospadoni ?
Thanks for your time :)

edospadoni:
It looks great. Nice work @federicoballarini :clap:

gsanchietti:
I agree, very very nice!

PR #175, opened 2019-11-27
ui. remove overlay on page change

federicoballarini:
This remove overlay when you change page and a modal still open.
Refer to https://github.com/NethServer/dev/issues/5961


edospadoni:
You have to catch the page change in SoftwareCenter.vue file and add the "modal hide" action on the hook. You can find many examples in others components like Backup.vue. The snipper is like this:

```js
beforeRouteLeave(to, from, next) {
    $(".modal").modal("hide");
    next();
  },
```

PR #176, opened 2019-11-28
ui. improve modal view

federicoballarini:
Changed into network, services and backup.
Also added `modal-lg` class for service status, backup log and certificate show.
Refer to https://github.com/NethServer/dev/issues/5965

### DIFF
--- a/ui/src/components/system/Backup.vue
+++ b/ui/src/components/system/Backup.vue
@@ -749,7 +749,7 @@
                 class="btn btn-default"
                 type="button"
                 data-dismiss="modal"
-              >{{$t('cancel')}}</button>
+              >{{$t('close')}}</button>
             </div>
           </form>
         </div>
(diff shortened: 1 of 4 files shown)


edospadoni:
There is no need to add `modal-lg` class inside modal. This PR must only change the labels.

federicoballarini:
@edospadoni Changed. But a question: you don't want to change modals size or I have to split it into another PR? Because in these 3 cases it can be helpful to have `modal-lg` size (backup logs, service status, show certificate).

edospadoni:
> @edospadoni Changed. But a question: you don't want to change modals size or I have to split it into another PR? Because in these 3 cases it can be helpful to have `modal-lg` size (backup logs, service status, show certificate).

Could be useful, but we don't have any `modal-lg` inside cockpit core or modules. We can evaluate a restyling in all modules, but next time.

federicoballarini:
Except the new Software Center ^_^

PR #178, opened 2019-12-01
ui. show dhcp lease expiration

federicoballarini:
Add lease expiration column on DHCP page.
Refer to: https://github.com/NethServer/dev/issues/5972

### DIFF
--- a/ui/src/components/system/DHCP.vue
+++ b/ui/src/components/system/DHCP.vue
@@ -998,7 +998,7 @@ export default {
                   Description: "",
                   IpAddress: lease.ip,
                   MacAddress: lease.mac,
-                  LeaseExpiration: new Date(lease.expire*1000).toLocaleString(navigator.language, {dateStyle: "medium", timeStyle: "medium", hour12: false, timeZone: Intl.DateTimeFormat().resolvedOptions().timeZone}).split(",").join(" ")
+                  LeaseExpiration: new Date(lease.expire*1000).toLocaleString(context.$i18n.locale, {dateStyle: "medium", timeStyle: "medium", hour12: false, timeZone: Intl.DateTimeFormat().resolvedOptions().timeZone}).split(",").join(" ")
                 },
                 type: "disabled"
               });
--- a/ui/src/components/system/DHCP.vue
+++ b/ui/src/components/system/DHCP.vue
@@ -100,7 +100,7 @@
           {{props.row.props.MacAddress}}
         </td>
         <td class="fancy">
-          <span v-show="props.row.props.LeaseExpiration" class="fa fa-hourglass-end margin"></span>
+          <span class="fa fa-hourglass-end margin"></span>
           {{props.row.props.LeaseExpiration | dateFormat}}
         </td>
         <td>
@@ -689,7 +689,8 @@ export default {
         {
           label: this.$i18n.t("dhcp.lease_expiration"),
           field: "props.LeaseExpiration",
-          filterable: true
+          filterable: true,
+          type: "number"
         },
         {
           label: this.$i18n.t("action"),
--- a/ui/src/components/system/DHCP.vue
+++ b/ui/src/components/system/DHCP.vue
@@ -92,16 +92,16 @@
         </td>
         <td class="fancy">{{ props.row.props.Description}}</td>
         <td class="fancy">
-          <span class="fa fa-desktop" style="margin-right: 5px"></span>
+          <span class="fa fa-desktop margin"></span>
           {{props.row.props.IpAddress}}
         </td>
         <td class="fancy">
-          <span class="pficon pficon-plugged" style="margin-right: 5px"></span>
+          <span class="pficon pficon-plugged margin"></span>
           {{props.row.props.MacAddress}}
         </td>
         <td class="fancy">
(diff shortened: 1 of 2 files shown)


federicoballarini:
Output should be something like this:

Schermata 2019-12-01 alle 15 19 32

ui/src/components/system/DHCP.vue
```diff
@@ -92,13 +92,17 @@
...
           {{props.row.props.IpAddress}}
         </td>
         <td class="fancy">
-          <span class="pficon pficon-plugged"></span>
+          <span class="pficon pficon-plugged" style="margin-right: 5px"></span>
           {{props.row.props.MacAddress}}
         </td>
+        <td class="fancy">
+          <span v-show="props.row.props.LeaseExpiration" class="fa fa-hourglass-end" style="margin-right: 5px"></span>
+          {{props.row.props.LeaseExpiration}}
```
edospadoni: add `dateFormat` filter here:

`{{props.row.props.LeaseExpiration | dateFormat}}`

ui/src/components/system/DHCP.vue
```diff
@@ -988,7 +997,8 @@ export default {
                 props: {
                   Description: "",
                   IpAddress: lease.ip,
-                  MacAddress: lease.mac
+                  MacAddress: lease.mac,
+                  LeaseExpiration: new Date(lease.expire*1000).toLocaleString(context.$i18n.locale, {dateStyle: "medium", timeStyle: "medium", hour12: false, timeZone: Intl.DateTimeFormat().resolvedOptions().timeZone}).split(",").join(" ")
```
edospadoni: remove all Date formatting and keep only:

`LeaseExpiration: lease.expire`

ui/src/components/system/DHCP.vue
```diff
@@ -92,13 +92,17 @@
         </td>
         <td class="fancy">{{ props.row.props.Description}}</td>
         <td class="fancy">
-          <span class="fa fa-desktop"></span>
+          <span class="fa fa-desktop" style="margin-right: 5px"></span>
```
edospadoni: remove inline style, add an existing class or create new one

ui/src/components/system/DHCP.vue
```diff
@@ -92,13 +92,17 @@
...
           {{props.row.props.IpAddress}}
         </td>
         <td class="fancy">
-          <span class="pficon pficon-plugged"></span>
+          <span class="pficon pficon-plugged" style="margin-right: 5px"></span>
```
edospadoni: remove inline style, add an existing class or create new one

federicoballarini:
Changed. It should be ok now.

federicoballarini:
There is a problem with sorting. I think it's caused by `-` that returns `dateFormat` if no date is found.
I think we can change line 104 into

`<span v-show="props.row.props.LeaseExpiration">{{props.row.props.LeaseExpiration | dateFormat}}</span>`

What do you think? @edospadoni

edospadoni:
> There is a problem with sorting. I think it's caused by `-` returned by `dateFormat` if no date is found.
> I think we can change line 104 into
> 
> `<span v-show="props.row.props.LeaseExpiration">{{props.row.props.LeaseExpiration | dateFormat}}</span>`
> 
> What do you think? @edospadoni

The problem is on sorting function. You have to specify field type, in this case `number`, so add:

```js
type: "number"
```

in column definition

ui/src/components/system/DHCP.vue
```diff
@@ -682,6 +686,11 @@ export default {
           field: "props.MacAddress",
           filterable: true
         },
+        {
+          label: this.$i18n.t("dhcp.lease_expiration"),
+          field: "props.LeaseExpiration",
+          filterable: true
```
edospadoni: add `type: "number"` here

ui/src/components/system/DHCP.vue
```diff
@@ -92,13 +92,17 @@
...
+          <span class="fa fa-desktop margin"></span>
           {{props.row.props.IpAddress}}
         </td>
         <td class="fancy">
-          <span class="pficon pficon-plugged"></span>
+          <span class="pficon pficon-plugged margin"></span>
           {{props.row.props.MacAddress}}
         </td>
+        <td class="fancy">
+          <span v-show="props.row.props.LeaseExpiration" class="fa fa-hourglass-end margin"></span>
```
edospadoni: remove `v-show="props.row.props.LeaseExpiration"`

ui/src/components/system/DHCP.vue
federicoballarini: Yes, you're right! Code genius :)

ui/src/components/system/DHCP.vue
federicoballarini: Why do you want to remove it? It will show the hourglasses even if there is not lease expiration.

edospadoni: If there is no `lease expiration` a dash (`-`) is shown but the hourglass is always present to indicat the type of the field. But if you prefer to remove it no problem is only a cosmetis.

federicoballarini: Yes, ok. We can leave it shown. I proceed to update my PR.
Thanks.

PR #180, opened 2019-12-03
ui. sort users and groups

federicoballarini:
Add sorting function to sort users and groups on Cockpit interface.
Refer to https://github.com/NethServer/dev/issues/5969

### DIFF
--- a/ui/src/components/system/UsersGroups.vue
+++ b/ui/src/components/system/UsersGroups.vue
@@ -2398,7 +2398,7 @@ export default {
         null,
         function(success) {
           try {
-            success = JSON.parse(success);
+            success = context.sortList(JSON.parse(success));
           } catch (e) {
             console.error(e);
           }
@@ -2423,7 +2423,7 @@ export default {
         null,
         function(success) {
           try {
-            success = JSON.parse(success);
+            success = context.sortList(JSON.parse(success));
           } catch (e) {
             console.error(e);
           }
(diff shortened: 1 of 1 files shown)


ui/src/components/system/UsersGroups.vue
```diff
@@ -2436,6 +2436,27 @@ export default {
...
+        }
+      }
+
+      sortable.sort(function (a, b) {
+        var x = a[0].toLowerCase(),
+            y = b[0].toLowerCase();
+        return x < y ? -1 : x > y ? 1 : 0;
+      });
+      
+      var json = JSON.stringify(sortable);
```
gsanchietti: :fearful: 

You should never manipolate a string to convert one data structure to another.
Please bear in mind that an object doesn't have the concept of order: you need to use and array to preserve the sorting.

edospadoni: Yes, I agree with @gsanchietti. You have to declare a new object like this:
```js
var newUserObj = {};
Object.keys(oldUserObject).sort().forEach(function(key) {
  newUserObj[key] = oldUserObject[key];
});
```

than in `newUserObj` should have the object keys ordered.

federicoballarini: Sorry :D I made changes. Easier than it looks. Thank you so much.

ui/src/components/system/UsersGroups.vue
```diff
@@ -2398,7 +2398,7 @@ export default {
         null,
         function(success) {
           try {
-            success = JSON.parse(success);
+            success = sortList(JSON.parse(success));
```
edospadoni: missing `context` here:

`success = context.sortList(JSON.parse(success));`

ui/src/components/system/UsersGroups.vue
```diff
@@ -2423,7 +2423,7 @@ export default {
...
```
edospadoni: missing `context` here:

`success = context.sortList(JSON.parse(success));`

ui/src/components/system/UsersGroups.vue
federicoballarini: Yes, I saw it :) I will update in a minute. My bad

PR #181, opened 2019-12-06
Avoid certificate update event cut off

DavidePrincipi:
The `nethserver-cockpit-conf` action forcibly restart `cockpit-ws` causing the user session to be disconnected.

The disconnection occurs 
- during `nethserver-cockpit` RPM  updates (from the Applications page, or yum-cron and other YUM update runs)
- during the `certificate-update` event, causing the complete event to fail

https://github.com/NethServer/dev/issues/5977

### DIFF
--- a/ui/public/i18n/language.json
+++ b/ui/public/i18n/language.json
@@ -1114,7 +1114,7 @@
     "network": "Network",
     "tls_policy": "TLS Policy",
     "trusted_networks": "Trusted Networks",
-
+    "certificates": "Certificates",
     "sftp_password": "The password will not be saved. SSH connection will use a RSA key.",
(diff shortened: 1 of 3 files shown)


ui/public/i18n/language.json
```diff
@@ -359,6 +359,10 @@
     "add_domain": "Add domain",
     "renew_days": "Renew days",
     "request": "Request",
+    "cockpit_reload_warning_title": "Notice.",
+    "cockpit_reload_warning_message": "The new certificate is applied to the Cockpit Server Manager 90 seconds after the last user is disconnected.",
```
filippocarletti: ...last user disconnects.

DavidePrincipi: Fixed after merge

PR #182, opened 2019-12-08
ui.User-Group Fix scrollbar in user list

stephdl:
the user list display a scroll bar
Screenshot_2019-12-08 Users Groups - ns7loc14 nethservertest org

after my fix

Screenshot_2019-12-08 Users Groups - ns7loc14 nethservertest org(1)


edospadoni:
This should be ok but check also in mobile version if it is possibile to scroll and also if popup on *three dots menù* is shown correctly

stephdl:
Screenshot_2019-12-09 Users Groups - ns7loc10 nethservertest org(3)
Screenshot_2019-12-09 Users Groups - ns7loc10 nethservertest org(2)

mobile view of FF, for me it is workable @edospadoni

edospadoni:
Yes, it works as expected

PR #183, opened 2019-12-09
ui. show webdav backup engine options

federicoballarini:
After selected WebDAV as backup destination, backup engine options weren't available.
Refer to https://github.com/NethServer/dev/issues/5981


edospadoni:
Yes, the fix is ok. Good job @federicoballarini  :+1:

PR #189, opened 2019-12-11
Fix confirm password field reset

DavidePrincipi:
Ensure the password confirmation field is cleared when the change password dialog is opened, even after multiple times.

See also https://community.nethserver.org/t/cockpit-user-password-change-problem-no-problem-in-nethgui/14006

Partial fix of https://github.com/NethServer/dev/issues/5984


DavidePrincipi:
Thank you I will merge this PR after adding another one to fix the server-side validation.

PR #191, opened 2019-12-12
Two factors of authentication in cockpit

stephdl:
The PR  will allow the user to protect the authentication with a double factor, a secret is stored under /var/lib/nethserver/secrets/user@domain.name (hex key) and we create a `secret` to make a  otpauth url (this kind of URL `otpauth://totp/root@ns7loc11.nethservertest.org?secret=IQM53USEFGBL57JBWUG4AZD3FHIOTOJG`). This url is used by the JS library QrcodeVue to display the QR code in the settings page from that URL

Once the user has flashed the QR code, we force the user to read the token on his mobile and validate the input in the form, if the token is valid then the user can save the 2FA, the JS library otplib is used for this. If the user disable the 2FA the secret is removed from /var/lib/nethserver/ and a new one is regenerated.

we propose 5 recovery code, these PIN can be used one time

https://github.com/NethServer/dev/issues/6085

======= below obsoleted, we keep it for history ======

It is a POC, it should be obviously enhanced

`yum install nethserver-directory`

then install nethserver-cockpit from PR
the root user gets its code inside `/var/lib/nethserver/secrets/root`
 
install gen-oath-safe and run 

`gen-oath-safe root totp  $(cat /var/lib/nethserver/secrets/root)`

then flash the QR code with freeOTP or google authenticator

go to cockpit and use root to login, the code is changed each 30 seconds

https://community.nethserver.org/t/sftp-server-improvements/14246/16

to create a HOTP (6 unic code to recovery login), install  oathtool, then create the code
`oathtool -w 5 $(cat /var/lib/nethserver/secrets/root)`

to create the secret for the QRcode

```
oathtool -v -d6 $(cat /var/lib/nethserver/secrets/root)
Hex secret: de32ce972ee4530bf5bcd71b4033519628b7d5b2
Base32 secret: 3YZM5FZO4RJQX5N424NUAM2RSYULPVNS
Digits: 6
Window size: 0
Start counter: 0x0 (0)
```
the url must be : `otpauth://totp/root@ns7loc11.nethservertest.org?secret=3YZM5FZO4RJQX5N424NUAM2RSYULPVNS`

put inside `/etc/liboath/users.oath`

```
HOTP/T30	root	-	2ce55e1dff5bc4d9fc2803269ad898627262de40	0	523858	2020-03-02T16:44:27L
HOTP	root	-	2ce55e1dff5bc4d9fc2803269ad898627262de40	1	773561	2020-03-02T16:43:40L
```


DavidePrincipi:
Added wiki page to draft this new feature:

https://github.com/NethServer/dev/wiki/One-Time-Password-(OTP)-authentication

DavidePrincipi:
The PR wants to add many parts and their implementation is out of my knowledge. I need some background information to be on the same page. Can you give me some pointers?

Could you update the PR description by summarizing the purpose of the new config files and their expected contents?

stephdl:
maybe it is a bit early to talk on it, let me finish a bit more

For now I want just a switch to enable the QRcode in the UI and the template behind

api/system-settings/getOtpToken
```diff
@@ -0,0 +1,70 @@
...
+my $db = esmith::ConfigDB->open_ro;
+my $input = NethServer::ApiTools::readInput();
+my $username = $input->{'username'} || '';
+my $OtpStatus;
+my $key;
+my $domainName = $db->get_value('DomainName');
+my $systemName = $db->get_value('SystemName');
+
+my $file = '/var/lib/nethserver/2fa/'.$username;
+if (! $file || ! -f $file ) {
```
gsanchietti: The `file` variable can't be empty, you just initialized it ;)
```suggestion
if (! -f $file ) {
```

root/etc/e-smith/templates/etc/nethserver/otp.secret/10base
```diff
@@ -0,0 +1,25 @@
+{
+  use strict;
+  use warnings;
+  use esmith::ConfigDB;
+  use NethServer::Password;
```
gsanchietti: You don't need to open the ConfigDB: you already have access to it.
Also remove stricts and warnings which prevent normal template usage.

```suggestion
  use NethServer::Password;
```

root/etc/e-smith/templates/etc/nethserver/otp.secret/10base
```diff
@@ -0,0 +1,25 @@
...
+
+  my $db = esmith::ConfigDB->open_ro();
+  my $DomainName = $db->get_value('DomainName');
```
gsanchietti:
```suggestion
```

root/etc/e-smith/templates/etc/nethserver/otp.force/10base
```diff
@@ -0,0 +1,17 @@
+{
+  use strict;
+  use warnings;
+  use esmith::ConfigDB;
+
+  my $db = esmith::ConfigDB->open_ro();
+  my $domain = $db->get_value('DomainName');
```
gsanchietti: Not needed, just remove it.
```suggestion
```

root/etc/e-smith/templates/etc/nethserver/otp.force/10base
```diff
@@ -0,0 +1,17 @@
...
+
+  foreach my $name (glob('/var/lib/nethserver/2fa/*')) {
+      $name =~ s|/var/lib/nethserver/2fa/||g;
+      $OUT .= "$name\n";
+      # we need to display user and user@domain.org
+      next if ($name eq 'root');
+      $name =~ s/\@$domain$//;
```
gsanchietti: use global variable available in all templates:
```suggestion
      $name =~ s/\@$DomainName$//;
```

api/system-settings/getOtpToken
stephdl: done

root/etc/e-smith/templates/etc/nethserver/otp.secret/10base
stephdl: done

root/etc/e-smith/templates/etc/nethserver/otp.force/10base
stephdl: done

root/etc/e-smith/templates/etc/nethserver/otp.force/10base
stephdl: done

root/etc/e-smith/templates/etc/nethserver/otp.secret/10base
stephdl: done

stephdl:
it should be good, I put the inline code, it is up to you to fill it @gsanchietti

stephdl:
We have an issue, it seems it is not possible to verify the token when the OS is windows, I think I know why, else I could have a workaround by using not the JS library but the oathtool CLI (and our standard API)

For arrows I will look but I set the field as numeric (to ease the verification), need to check if I can hide the arrow, still a noob web developer :D

PR #192, opened 2019-12-13
ui. fix files download on FF

federicoballarini:
Changed backup download method.
Refer to https://github.com/NethServer/dev/issues/5988

### DIFF
--- a/ui/src/components/system/Backup.vue
+++ b/ui/src/components/system/Backup.vue
@@ -68,10 +68,10 @@
           </div>
           <div class="list-group-item" v-for="(b, bk) in backupConfigurations" v-bind:key="bk">
             <div class="list-view-pf-actions">
-              <button @click="downloadConfigBackup(b)" class="btn btn-default">
+              <a :id="b.id + '-button'" class="btn btn-default">
                 <span class="fa fa-download span-right-margin"></span>
                 {{$t('backup.download')}}
-              </button>
+              </a>
               <div class="dropup pull-right dropdown-kebab-pf">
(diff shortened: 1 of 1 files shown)


federicoballarini:
@edospadoni I changed download method as `vpn-ui` but I get a CSP error on FF. CSP seems equal to `vpn-ui` one, but it doesn't work. Can you look if we can fix it?

edospadoni:
> @edospadoni I changed download method as `vpn-ui` but I get a CSP error on FF. CSP seems equal to `vpn-ui` one, but it doesn't work. Can you look if we can fix it?

I have no idea :smile: . It the same problem that I faced weeks ago when I have implemented the download with strange name

edospadoni:
> @edospadoni I changed download method as `vpn-ui` but I get a CSP error on FF. CSP seems equal to `vpn-ui` one, but it doesn't work. Can you look if we can fix it?

I have no idea :smile: . It the same problem that I faced weeks ago when I have implemented the download with strange name.

gsanchietti:
So, we should revert everything and close as wantfix?

federicoballarini:
No, I think I found a solution. I try a commit in a minute.

federicoballarini:
I'm working on it to find a solution. I will update you asap.

federicoballarini:
I can't find out any solution. I think we can close this PR and start testing only `vpn-ui`. Probably there are problems on Firefox, but also on Cockpit main module, because I can't understand why on `vpn-ui` it works fine.

PR #194, opened 2019-12-14
ui & api. add disk information on main dashboard

federicoballarini:
Added disk information on Cockpit Dashboard.
Refer to https://github.com/NethServer/dev/issues/5990


federicoballarini:
The result should be like this:

Schermata 2019-12-15 alle 01 04 24

If RAID is configured it will shows `RAID: OK` or `RAID: ERROR` into root partition box.
I copied raid-check script into `api/system-status` folder because I don't know if we want to keep the script in the old position (`/usr/libexec/nethserver/raid-status`)

edospadoni:
> If RAID is configured it will shows `RAID: OK` or `RAID: ERROR` into root partition box.
> I copied raid-check script into `api/system-status` folder because I don't know if we want to keep the script in the old position (`/usr/libexec/nethserver/raid-status`)

1. I think it's better to separate the RAID status from disk status, but I would like to see some screenshots to visualize the idea. Could you provide screenshot for RAID status, both ok and error state?

2. We have to keep `/usr/libexec/nethserver/raid-status` in the original position, the new file called `api/system-status/check-raid` can simply invoke the script and works like a wrapper. We have some examples of cockpit APIs that work only like a wrapper.

3. The UI part is ok, but I prefer to separate this: https://github.com/NethServer/nethserver-cockpit/pull/194/files#diff-e8d2bf840a9258188156d22211950a6dR65 from this PR

federicoballarini:
For point one these are two screenshot. If no raid is configured view is like the old one screenshot.

image

image

For point two and three I will commit changes after we decide what to do with RAID.

edospadoni:
It's not bad :thinking: I like it, but I prefer the keep the color of the`ERROR` label red, the standard one.

ui/src/components/system/Dashboard.vue
```diff
@@ -1761,4 +1853,7 @@ export default {
 </script>
 
 <style>
+  .raidError {
+    color: #470000;
```
edospadoni: change to `#cc0000`, but I think there is a `.red` CSS class inside the cockpit-core project that you can use

federicoballarini: Used red class.

api/system-status/read
```diff
@@ -32,7 +32,45 @@ function readMemory()
...
+        $ret[$tmp[5]] = array('total' => intval($tmp[1]), 'used' => intval($tmp[2]));
+    }
+    return $ret; 
+}
+
+function readRaid()
+{
+    $out = array();
+    $ret = array();
+    $out = json_decode(exec('/usr/libexec/nethserver/api/system-status/check-raid'), true);
```
DavidePrincipi: If `system-status/check-raid` is not used elsewhere, we can simplify and optimize the process calls:

```suggestion
    $out = json_decode(exec('/usr/libexec/nethserver/raid-status'), true);
```

federicoballarini: Yes, sure! I will do it. And I will remove check-raid file.

ui/src/components/system/Dashboard.vue
```diff
@@ -907,9 +954,21 @@ export default {
               available_bytes: success.status.memory.SwapTotal
             }
           };
+          
+          context.system.disk = {
+            root: {
+              used_bytes: success.status.disk.root.used,
+              available_bytes: success.status.disk.root.total
```
edospadoni: I think this line has to be changed to `success.status.disk.root.available` not `success.status.disk.root.total`

federicoballarini: Yes, right. It's a typo error. I fix it.

federicoballarini: @edospadoni I think there is the same problem with ram and swap. Chart is not correct because available_bytes is total size. I commit changes only for disk, tell me if I have to change behaivour also on ram and swap.

federicoballarini:
@DavidePrincipi remove check-raid file.
@edospadoni now disk is ok. Wait for your feedback about ram and swap.

Schermata 2019-12-17 alle 13 23 03

ui/src/components/system/Dashboard.vue
edospadoni: I don't think so :thinking:

The SWAP is correct:
```js
used_bytes:
    success.status.memory.SwapTotal -
    success.status.memory.SwapFree,
available_bytes: success.status.memory.SwapTotal
```

The RAM is correct:
```js
used_bytes:
    success.status.memory.MemTotal -
    success.status.memory.MemAvailable,
available_bytes: success.status.memory.MemTotal
```

Try to check the values with the command `free -h`

ui/src/components/system/Dashboard.vue
```diff
@@ -907,9 +955,22 @@ export default {
               available_bytes: success.status.memory.SwapTotal
             }
           };
+          
+          context.system.disk = {
+            root: {
+              used_bytes: success.status.disk.root.used,
+              available_bytes: success.status.disk.root.free,
+              total_bytes: success.status.disk.root.total
```
edospadoni: you can use only `used_bytes` and `available_bytes`. because the total is automatically calculated with the chart.

```js
used_bytes:
    success.status.disk.root.total -
    success.status.disk.root.free,
available_bytes: success.status.disk.root.total or success.status.disk.root.available (depends on your API)
```

federicoballarini: I used `total_bytes` for size span on the right bottom of the box. If you want I can remove `total_bytes` and perform a sum of used and available.
I didn't make a subtraction of `total` and `free` beacuse api can return also used disk space.

ui/src/components/system/Dashboard.vue
federicoballarini: image

If 2.01 GB are used of 3.7 total GB, the chart percentage should be over 50%. The error is that we use available_bytes as total in the chart column.

federicoballarini: If 2.01 GB are used of 3.7 total GB, the chart percentage should be over 50%. The error is that we use available_bytes as total in the chart column.

image

ui/src/components/system/Dashboard.vue
edospadoni: ok, leave it like this.

ui/src/components/system/Dashboard.vue
edospadoni: ok, only in the RAM card, SWAP seems right. You can fix also the RAM chart to match the correct value. thanks

federicoballarini: Can I commit here or do you prefer another PR?

edospadoni: Another PR, this PR is focused on disk card, this RAM fix is different, but you can link the same Issue.

federicoballarini: Opened PR here: https://github.com/NethServer/nethserver-cockpit/pull/197

api/system-status/read
```diff
@@ -73,10 +111,10 @@ function readRelease()
 }
 
 function readKernel()
-    {
```
DavidePrincipi: Please remove this whitespace-only change!

federicoballarini: Removed.

api/system-status/read
```diff
@@ -32,7 +32,45 @@ function readMemory()
...
+        $ret[$tmp[5]] = array('total' => intval($tmp[1]), 'used' => intval($tmp[2]), 'free' => intval($tmp[3]));
+    }
+    return $ret; 
+}
+
+function readRaid()
+{
+    $out = array();
+    $ret = array();
+    $out = json_decode(exec('/usr/libexec/nethserver/raid-status'), true);
```
gsanchietti: If you need just to report if there are any errors, you do not need to parse the output: just check the exit value.
If the exit value is different than 0, md is are errors.

federicoballarini: This is new for me ^-^

The output of an OK RAID is this:
```
[root@fw ~]# /usr/libexec/nethserver/raid-status
{"results":[{"md":"md1","tot_devs":"2","spare_devs":"none","active_devs":"sdb1,sda1","failed_devs":"none","ok_devs":"2","level":"1","result":"ok"},{"md":"md2","tot_devs":"2","spare_devs":"none","active_devs":"sdb2,sda2","failed_devs":"none","ok_devs":"2","level":"1","result":"ok"}]}
```

How I have to change my code?

gsanchietti: This is the magic line on the raid script: https://github.com/NethServer/nethserver-base/blob/master/root/usr/libexec/nethserver/raid-status#L100

To use it, just change your code like (I didn't test it):
```php
exec('/usr/libexec/nethserver/raid-status', $out, $exit_code);
if ($exit_code == 0) {
        $ret['status'] = 'ok';
} else {
        $ret['status'] = 'error';
}
return $ret
```

federicoballarini: I made some changes to hide RAID status if RAID is not configured. Can you check if it's ok for you? It should works fine.

api/system-status/read
```diff
@@ -32,7 +32,48 @@ function readMemory()
...
+    $records = count($out["results"]);
+    
+    if ($records > 0) {
+      if ($exit_code == 0) {
+        $ret['status'] = 'ok';
+      } else {
+        $ret['status'] = 'error';
+      }
+    } else {
+      $ret['status'] = false;
```
DavidePrincipi: I know it's just a matter of taste... I prefer `NULL` to represent a missing information. Here to represent a missing RAID device I'd use NULL instead of FALSE.

```suggestion
      $ret['status'] = NULL;
```

api/system-status/read
```diff
@@ -54,17 +54,20 @@ function readDisk()
...
 {
-    $out = array();
     $ret = array();
-    $out = json_decode(exec('/usr/libexec/nethserver/raid-status'), true);
-    $mds = $out['results'];
-    $ret['status'] = false;
-    for ($i=0; $i<count($mds) && $ret['status'] != 'error'; $i++) {
-      if ($mds[$i]['failed_devs'] == 'none') {
+    $out = array();
+    exec('/usr/libexec/nethserver/raid-status', $out, $exit_code);
```
gsanchietti: Now I get it: the `$ret['status']` variable is not a boolean, but has 3 different states so you need always to parse the JSON, just what I was trying to avoid ;)

federicoballarini: I will `change` false to `NULL`, also I prefer it.
I left code as the last commit if it's ok. I think that count number of records to understand if raid is enabled is ok.

api/system-status/read
federicoballarini: I love philosophical discussions ^-^
I will change it in a minute :)

PR #197, opened 2019-12-17
ui. fix memory and swap charts

federicoballarini:
Fixed memory and swap charts percentage.
Refer to https://github.com/NethServer/dev/issues/5990


federicoballarini:
image

From this image we can see that RAM and SWAP charts were wrong.

PR #201, opened 2019-12-28
api. limit interface on hotsync slave host

federicoballarini:
Limit interface pages if Cockpit is used on HotSync slave host.
Refer to https://github.com/NethServer/dev/issues/6014


gsanchietti:
The hotsync is an extra package and can't add any code to core.
This is the principle of modularity that we should not break.

Sorry but I have to reject this PR.

federicoballarini:
Ok, we only write into docs to not make updates form Cockpit GUI if not required from hotsync scripts.

PR #206, opened 2020-01-13
Group based permission access to cockpit

stephdl:
https://github.com/NethServer/dev/issues/6029

### DIFF
--- a/createlinks
+++ b/createlinks
@@ -42,6 +42,12 @@ event_templates('nethserver-cockpit-update', qw(
     /etc/pam.d/cockpit
 ));
 
+# expand cockpit PAM following `ShellOverrideStatus`
+event_templates('nethserver-sssd-save', qw(
+    /etc/pam.d/cockpit
+    /etc/nethserver/cockpit.allow
+));
+
 # delegate cockpit access via PAM
 event_templates('delegation-save', qw(
     /etc/nethserver/cockpit.allow
--- a/createlinks
+++ b/createlinks
@@ -38,6 +38,13 @@ event_templates('nethserver-cockpit-update', qw(
     /etc/nethserver/ldappasswd.conf
     /etc/sudoers.d/30_nethserver_cockpit_roles
     /etc/sudoers.d/55_nsapi_perms
+    /etc/nethserver/cockpit.allow
+    /etc/pam.d/cockpit
+));
+
+# delegate cockpit access via PAM
+event_templates('delegation-save', qw(
+    /etc/nethserver/cockpit.allow
 ));
 
 #
--- a/root/etc/e-smith/templates.metadata/etc/nethserver/cockpit.allow
+++ b/root/etc/e-smith/templates.metadata/etc/nethserver/cockpit.allow
@@ -1,3 +1,0 @@
-UID="root"
-GID="root"
-PERMS="0440"
--- a/root/etc/e-smith/templates.metadata/etc/nethserver/cockpit.allow
+++ b/root/etc/e-smith/templates.metadata/etc/nethserver/cockpit.allow
@@ -1,0 +1,3 @@
+UID="root"
+GID="root"
+PERMS="0440"
--- a/root/etc/e-smith/templates.metadata/etc/pam.d/cockpit
+++ b/root/etc/e-smith/templates.metadata/etc/pam.d/cockpit
@@ -1,0 +1,3 @@
+UID="root"
+GID="root"
+PERMS="0644"
--- a/root/etc/e-smith/templates.metadata/etc/pam.d/cockpit
+++ b/root/etc/e-smith/templates.metadata/etc/pam.d/cockpit
@@ -1,3 +1,0 @@
-UID="root"
-GID="root"
-PERMS="0644"
--- a/root/etc/e-smith/templates/etc/nethserver/cockpit.allow/10base
+++ b/root/etc/e-smith/templates/etc/nethserver/cockpit.allow/10base
@@ -1,0 +1,20 @@
+
+#
+# Allow groups to use cockpit 
+#
+
+root
+{ $admins{'group'} || 'domain admins'}
+
+{
+    foreach my $rule (split(',', ${'cockpit.socket'}{'delegation'} || '')) {
+
+        my ($group, @modules) = split(':', $rule);
+
+        if(!$group || !@modules) {
+            next;
+        }
+        
+        $OUT .= "$group\n";
+    }
+}
--- a/root/etc/e-smith/templates/etc/pam.d/cockpit/10base
+++ b/root/etc/e-smith/templates/etc/pam.d/cockpit/10base
@@ -3,24 +3,3 @@ auth	   required	pam_sepermit.so
 auth       substack     password-auth
 auth       include      postlogin
 auth       optional     pam_ssh_add.so
-{
-if (($sssd{'ShellOverrideStatus'} || 'disabled') eq 'enabled') {
-  $OUT .= q(
-# select group allowed to login to cockpit (root & domain admins alway OK)
-auth    required       pam_listfile.so \
-        onerr=fail  item=group  sense=allow  file=/etc/nethserver/cockpit.allow
-);
-}
-}
-account    required     pam_nologin.so
-account    include      password-auth
-password   include      password-auth
-# pam_selinux.so close should be the first session rule
-session    required     pam_selinux.so close
-session    required     pam_loginuid.so
-# pam_selinux.so open should only be followed by sessions to be executed in the user context
-session    required     pam_selinux.so open env_params
-session    optional     pam_keyinit.so force revoke
-session    optional     pam_ssh_add.so
-session    include      password-auth
-session    include      postlogin
--- a/root/etc/e-smith/templates/etc/pam.d/cockpit/10base
+++ b/root/etc/e-smith/templates/etc/pam.d/cockpit/10base
@@ -1,0 +1,26 @@
+#%PAM-1.0
+auth	   required	pam_sepermit.so
+auth       substack     password-auth
+auth       include      postlogin
+auth       optional     pam_ssh_add.so
+{
+if (($sssd{'ShellOverrideStatus'} || 'disabled') eq 'enabled') {
+  $OUT .= q(
+# select group allowed to login to cockpit (root & domain admins alway OK)
+auth    required       pam_listfile.so \
+        onerr=fail  item=group  sense=allow  file=/etc/nethserver/cockpit.allow
+);
+}
+}
+account    required     pam_nologin.so
+account    include      password-auth
+password   include      password-auth
+# pam_selinux.so close should be the first session rule
+session    required     pam_selinux.so close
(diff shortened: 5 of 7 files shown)


root/etc/e-smith/templates/etc/pam.d/cockpit/10base
```diff
@@ -0,0 +1,27 @@
...
+auth [success=1 default=ignore] pam_succeed_if.so user ingroup [domain admins]
+# select group allowed to login to cockpit (root & domain admins alway OK)
+auth    required       pam_listfile.so \
+        onerr=fail  item=group  sense=allow  file=/etc/nethserver/authorizations/cockpit.access
+account    required     pam_nologin.so
+account    include      password-auth
+password   include      password-auth
+# select group allowed to login to cockpit (root & domain admins alway OK)
+auth    required       pam_listfile.so \
+        onerr=fail  item=group  sense=allow  file=/etc/nethserver/authorizations/cockpit.access
```
DavidePrincipi: Is this a repeated line?

nethserver-cockpit.spec
```diff
@@ -40,6 +40,7 @@ perl createlinks
 
 %install
 (cd root ; find . -depth -not -name '*.orig' -print | cpio -dump %{buildroot})
+mkdir -p %{buildroot}/etc/nethserver/authorizations/
```
DavidePrincipi: I'd like to avoid a new directory... Let's put the pam_listfile in `/etc/nethserver/cockpit.allow` or similar.

root/etc/e-smith/templates/etc/pam.d/cockpit/10base
stephdl: yes, truncated, you prefer one line ?

DavidePrincipi: No, the line break is fine. I wanted to say that the above lines 17-18 are equal to 11-12. They seem repeated, and should be removed IMO.

stephdl: oups got it, 16-17-18 must be removed it cannot work

stephdl: done

nethserver-cockpit.spec
stephdl: done

root/etc/e-smith/templates/etc/pam.d/cockpit/10base
```diff
@@ -0,0 +1,24 @@
+#%PAM-1.0
+auth	   required	pam_sepermit.so
+auth       substack     password-auth
+auth       include      postlogin
+auth       optional     pam_ssh_add.so
+# the root and members of domain admins are always allowed
+# if pam auth succeed, it ignores  pam_listfile.so
+auth [success=2 default=ignore] pam_succeed_if.so user in [root]
+auth [success=1 default=ignore] pam_succeed_if.so user ingroup [domain admins]
+# select group allowed to login to cockpit (root & domain admins alway OK)
```
DavidePrincipi: I think we can remove the above lines and implement the same rule in `cockpit.allow` template logic.

root/etc/e-smith/templates/etc/pam.d/cockpit/10base
```diff
@@ -0,0 +1,24 @@
...
+auth    required       pam_listfile.so \
```
DavidePrincipi: This module is not always required. If the ShellOverrideStatus prop is disabled, we must comment it to implement the "legacy" behavior. 

I'd prefer to not output it at all, so that our template echoes out the same original PAM config from the cockpit RPM.

root/etc/e-smith/templates/etc/nethserver/cockpit.allow/10base
```diff
@@ -0,0 +1,17 @@
+
+#
+# Allow groups to use cockpit 
+#
+
+{
```
DavidePrincipi: We can always list `root` and `domain admins` here. The file has to be checked by PAM only if `ShellOverrideStatus=enabled`.

root/etc/e-smith/templates/etc/pam.d/cockpit/10base
stephdl: well if something goes wron in the template expansion ,then root and doamin admins are locked out ....it could be dangerous

I think that root and domain admins group must be accepted all the time

root/etc/e-smith/templates/etc/nethserver/cockpit.allow/10base
stephdl: of course but you might have a chance that the template goes wrong

DavidePrincipi: What could go wrong with this template? `/etc/pam.d/cockpit` is a template too...

stephdl: ok

root/etc/e-smith/templates.metadata/etc/nethserver/cockpit.allow
```diff
@@ -0,0 +1,3 @@
+UID="root"
+GID="root"
+PERMS="0440"
```
DavidePrincipi: Why such restrictive permissions? Is't the default root/root `rw-r--r--` ok?

I think this metadata file is not required.

root/etc/e-smith/templates.metadata/etc/pam.d/cockpit
```diff
@@ -0,0 +1,3 @@
+UID="root"
+GID="root"
+PERMS="0644"
```
DavidePrincipi: I think this metadata file is not required. Why do you want to add it?

root/etc/e-smith/templates.metadata/etc/nethserver/cockpit.allow
stephdl: yes this one is on my own, maybe we could be less strict :-?

root/etc/e-smith/templates.metadata/etc/pam.d/cockpit
stephdl: yes this one is on my own, maybe we could be less strict :-?

root/etc/e-smith/templates/etc/pam.d/cockpit/10base
stephdl: done

root/etc/e-smith/templates.metadata/etc/nethserver/cockpit.allow
stephdl: done

root/etc/e-smith/templates.metadata/etc/pam.d/cockpit
stephdl: done

createlinks
```diff
@@ -38,6 +38,21 @@ event_templates('nethserver-cockpit-update', qw(
...
+));
+
+# expand cockpit PAM following `ShellOverrideStatus`
+event_templates('nethserver-sssd-save', qw(
+    /etc/pam.d/cockpit
+));
+
+# delegate cockpit access via PAM
+event_templates('group-modify', qw(
+    /etc/nethserver/cockpit.allow
```
DavidePrincipi: I can't figure out a use case where this is needed, both for `group-modify` and `group-create`. Instead the `group-delete` could be interesting. But it would work only with a local account provider.

stephdl: you are right we miss the group delete, we have to maintain a list following the permission stored under ${cockpit.socket}{delegation}, so with group-{create,delete,modify} it entends to add or remove permission to login to cockpit

For a remote account provider, it is out of the scope of this issue, but our group panel should allow to add permission locally even if the group is remote

stephdl: an alternative is to use `delegation-save`. I will commit to use it

DavidePrincipi: :thinking: 

1. group-create: during this event no delegation occurs, for sure. 

2. group-modify: here too

3. group-delete: it _could_ update the `delegation` prop but it only works with a local accounts provider. Furthermore, if a group does not exist I think there is no problem because no user can match PAM/sshd rules with it.

Instead, we'd bind the `delegation-save` event.

stephdl: done we expand the list of group with `delegation-save`

createlinks
```diff
@@ -38,6 +38,18 @@ event_templates('nethserver-cockpit-update', qw(
     /etc/nethserver/ldappasswd.conf
     /etc/sudoers.d/30_nethserver_cockpit_roles
     /etc/sudoers.d/55_nsapi_perms
+    /etc/nethserver/cockpit.allow
+    /etc/pam.d/cockpit
+));
+
+# expand cockpit PAM following `ShellOverrideStatus`
+event_templates('nethserver-sssd-save', qw(
+    /etc/pam.d/cockpit
```
DavidePrincipi: Expanding also the cockpit.allow template does not waste. Use case: 

1. configure a remote account provider with custom `admins/group` prop
2. enable `ShellOverrideStatus`
3. run `nethserver-sssd-save`

```suggestion
    /etc/pam.d/cockpit
    /etc/nethserver/cockpit.allow
```

See also the next comment

root/etc/e-smith/templates/etc/nethserver/cockpit.allow/10base
```diff
@@ -0,0 +1,19 @@
+
+#
+# Allow groups to use cockpit 
+#
+root
+domain admins
```
DavidePrincipi: Here we'd expand the `admins/group` prop too...
```suggestion
{ $admins{'group'} }
```

createlinks
stephdl: done

root/etc/e-smith/templates/etc/nethserver/cockpit.allow/10base
stephdl: done

root/etc/e-smith/templates/etc/nethserver/cockpit.allow/10base
```diff
@@ -0,0 +1,19 @@
+
+#
+# Allow groups to use cockpit 
+#
+root
+{ $admins{'group'} || 'domain admins'}
```
stephdl: @davidep what about if the prop gets several group, in theory it is possible ?

Maybe we should split it ?

DavidePrincipi: The `group` prop is singular, so only _one_ group can be set.

https://docs.nethserver.org/en/v7/accounts.html#admin-account

Can I have seen some `split()` on `group` somwhere else? Please fix it!

stephdl: yep in my code inside openssh

stephdl: done, thank you

PR #209, opened 2020-01-20
Hide shell checkbox in case of forced shell

stephdl:
https://github.com/NethServer/dev/issues/6029


DavidePrincipi:
It was Steph's first choice. Then we discussed about it and decided to remove it completely because in similar situations we preferred to hide over disabling. The disabled state may be misleading: the user might say "why I cant change this checkbox"?

We must document in 7.8 release notes (and possibly in related sections) how the ShellOverrideStatus prop changes the behavior of the system and the UI.

gsanchietti:
Ok fine, let's document it better in the linked manual page.

PR #211, opened 2020-01-20
Do not delegate logs

stephdl:
https://github.com/NethServer/dev/issues/6033


DavidePrincipi:
There could be also another way to implement the root-only access to the System/Logs page

https://github.com/NethServer/nethserver-cockpit/blob/49afeaeabe370341a87e76d9a184e43d0575f5d9/ui/src/App.vue#L204-L214

Let's discuss pros/cons of each approach

stephdl:
obsoleted by https://github.com/NethServer/nethserver-cockpit/pull/213

PR #212, opened 2020-01-24
ui & api. added yum clean button

edospadoni:
The `yum clean all` action is added into the Software Center page with a primary button with label `Update`. Screenshot attached.

yum_cache


filippocarletti:
Still undecided between Refresh and Reload. Need to flip a coin. :-)

PR #215, opened 2020-02-03
api: prevent shell injection

gsanchietti:
NethServer/dev#6042

### DIFF
--- a/api/lib/users_functions.php
+++ b/api/lib/users_functions.php
@@ -25,19 +25,19 @@ require_once("/usr/libexec/nethserver/api/lib/Helpers.php");
 
 /* Return the groups of a user */
 function user_membership($user) {
-    $output = shell_exec("/usr/libexec/nethserver/list-user-membership -s '$user'");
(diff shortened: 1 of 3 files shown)


api/lib/users_functions.php
```diff
@@ -25,19 +25,19 @@
 
 /* Return the groups of a user */
 function user_membership($user) {
-    $output = shell_exec("/usr/libexec/nethserver/list-user-membership -s '$user'");
+    $output = shell_exec(escapeshellcmd("/usr/libexec/nethserver/list-user-membership -s ".escapeshellarg($user)));
```
DavidePrincipi: `escapeshellcmd()` should not be needed because the inner `escapeshellarg()` does all the job.

```suggestion
    $output = shell_exec("/usr/libexec/nethserver/list-user-membership -s " . escapeshellarg($user));
```

I'd call `escapeshellcmd()` only on a whole shell command line where individual arguments are not under my control. That's not what happens here: our `$user` variable corresponds to an argument and is the only argument to protect.

PR #226, opened 2020-03-02
Fix Save button lockout

DavidePrincipi:
Avoid saving an empty value in the props that always require a value.

Strange but it can happen if the UI is not up-to-date with the underlying API.

See

https://github.com/NethServer/dev/issues/6075

### DIFF
--- a/api/system-openssh/update
+++ b/api/system-openssh/update
@@ -35,8 +35,8 @@ foreach my $prop (qw(
     TCPPort
     AllowEveryone
     )) {
-    my $value = $data->{'props'}{$prop};
-    if($value) {
+    my $value = $data->{'props'}{$prop} || '';
+    if($value ne '') {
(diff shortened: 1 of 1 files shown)


stephdl:
I bet that fix must be applied in other modules :-?

DavidePrincipi:
I share your concerns. This little bug/regression is typical with updates, as I stated here

https://community.nethserver.org/t/ssh-unable-to-save-changes-from-cockpit-ui/14808/27?u=davidep

Maybe it requires to reconsider the way we apply updates. However the fix works at least in this case and it avoids locking the SSH interface and maybe calls to support.

gsanchietti:
I think the fix is ok: the API should take care to avoid error on saving.
Eventually, also a robust template can handle wrong properties config.

stephdl:
I have a double mind, the fix avoid to save empty properties, it doesn't fix the issue.

PR #227, opened 2020-03-02
AllowGroups must be an object

stephdl:
https://github.com/NethServer/dev/issues/6075

### DIFF
--- a/api/system-openssh/update
+++ b/api/system-openssh/update
@@ -42,15 +42,19 @@ foreach my $prop (qw(
 }
 
 my $AllowGroups = '';
-foreach my $g (keys $data->{'props'}{'AllowGroups'} ) {
-    if ($data->{'props'}{'AllowGroups'}{$g} eq 'ssh') {
-        $AllowGroups .= "$g,";
-    } elsif ($data->{'props'}{'AllowGroups'}{$g} eq 'sftp') {
-        $AllowGroups .= "$g:sftp,";
-    }
-} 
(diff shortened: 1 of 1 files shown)


stephdl:
the AllowGroups must be an object because we use the `keys`

https://github.com/NethServer/nethserver-cockpit/blob/master/api/system-openssh/update#L45

api/system-openssh/update
```diff
@@ -42,19 +42,16 @@ foreach my $prop (qw(
...
+foreach my $g (keys $obj) {
+    if ($obj->{$g} eq 'ssh') {
+        $AllowGroups .= "$g,";
+    } elsif ($obj->{$g} eq 'sftp') {
+        $AllowGroups .= "$g:sftp,";
+    }
 }
-
+chop $AllowGroups;
+$cdb->set_prop('sshd', 'AllowGroups', $AllowGroups) if(ref($data->{'props'}{'AllowGroups'}) eq 'HASH');
```
DavidePrincipi:
```suggestion
$cdb->set_prop('sshd', 'AllowGroups', $AllowGroups) if(ref($obj) eq 'HASH');
```

stephdl: it is merged, do you want to open a new PR ?

DavidePrincipi: I already pushed the commit to master :innocent:

stephdl: super

PR #231, opened 2020-03-18
2FA Without sudo

stephdl:
This version stores the 2FA key under the user permission, to avoid to use the cockpit API and retrieve the key of another user

NethServer/dev#6085


api/system-authorization/read
```diff
@@ -96,6 +98,6 @@ print encode_json({
     "status" => {
         "isRoot" => $isRoot,
         "isAdmin" => $isAdmin,
-        "username" => scalar getpwuid $<
+        "username" => $username
```
gsanchietti: This was present also on the previous PR.
Where is the new `username` parameter needed?

api/system-settings/getOtpToken
```diff
@@ -23,25 +23,34 @@
 use warnings;
```
gsanchietti: Cosmetic: please do not use filename with camel case.
See all other helpers. The name should look like: get-otp-token

api/system-settings/getOtpToken
```diff
@@ -23,25 +23,34 @@
...
 use JSON;
-use esmith::ConfigDB;
 use NethServer::ApiTools;
 use Convert::Base32;
 use NethServer::Password;
+use Sys::Hostname;
 
 my $db = esmith::ConfigDB->open_ro;
 my $input = NethServer::ApiTools::readInput();
 my $username = $input->{'username'} || '';
```
gsanchietti: Why the cockpit is calling this script directly?? From the browser console:
```
 echo '{"username":"giacomo"}' | /usr/libexec/nethserver/api/system-settings/getOtpToken | jq
```
This is against cockpit design: you should always call an api.

Also since the call is done without sudo, it doesn't make sense to read the user name from the stdin, just retrieve it from the script with something like:
```
my ($username,$passwd,$uid,$gid, $quota,$comment,$gcos,$home,$shell,$expire) = getpwuid( $< );

```

api/system-authorization/read
stephdl: above

gsanchietti: The `username` paramater is passed to the UI. What is for?
Do you just need to access the current logged user name from the UI?

stephdl: the user's name is needed to find the path of the user's home

stephdl:
I need to amen the QA

PR #234, opened 2020-03-24
Protect sshd with 2FA

stephdl:
https://github.com/NethServer/dev/issues/6088


ui/public/i18n/language.json
```diff
@@ -841,6 +841,7 @@
     "validate_the_QRcode_token": "Verification code",
     "test_token": "Check code",
     "OtpCockpit_status": "Enable 2FA for the Server Manager",
+    "OtpSshd_status": "Enable 2FA for SSH (password only)",
```
gsanchietti:
```suggestion
    "OtpSshd_status": "Enable 2FA for SSH",
```

ui/src/components/system/Settings.vue
```diff
@@ -243,6 +243,29 @@
...
           </div>
         </div>
+        <div
+          v-if="view.otpIsLoaded && otp.OtpStatus"
+          :class="['form-group', errors.OtpSshd.hasError ? 'has-error' : '']"
+        >
+          <label
+            class="col-sm-2 control-label"
+            for="textInput-modal-markup"
+          >{{$t('settings.OtpSshd_status')}}</label>
```
gsanchietti: Can we add an inline-doc here?
The tool tip should state something like:
```
2FA is available only for password authentication.
Authentication with public keys will not be affected by 2FA.
```

stephdl: @gsanchietti done

ui/public/i18n/language.json
stephdl: done

PR #239, opened 2020-03-27
Expose change password page to port 443

edospadoni:
https://github.com/NethServer/dev/issues/6074
https://github.com/NethServer/dev/issues/6100


root/etc/e-smith/templates/etc/httpd/conf.d/default-virtualhost.inc/15user_settings
```diff
@@ -0,0 +1,21 @@
...
+RewriteRule /(.*)           http://127.0.0.1:9090/$1 [P,L]
+
+# enable proxypass for ws and http
+ProxyPass /cockpit/socket ws://127.0.0.1:9090/cockpit/socket
+ProxyPassReverse /cockpit/socket ws://127.0.0.1:9090/socket
+ProxyPass /cockpit http://127.0.0.1:9090/cockpit
+ProxyPassReverse /cockpit http://127.0.0.1:9090
+
+# create redirect path to user settings
+Redirect "/user-settings" "/cockpit/@localhost/nethserver/index.html#/settings"
```
gsanchietti: Apache syntax is broken:
```suggestion
Redirect "/user-settings" "/cockpit/@localhost/nethserver/index.html#/settings"

```

createlinks
```diff
@@ -39,6 +39,8 @@ event_templates('nethserver-cockpit-update', qw(
     /etc/sudoers.d/30_nethserver_cockpit_roles
     /etc/sudoers.d/55_nsapi_perms
     /etc/nethserver/cockpit.allow
+    /etc/cockpit/cockpit.conf
+    /etc/httpd/conf.d/default-virtualhost.inc
 ));
```
gsanchietti:
```suggestion

event_services(nethserver-cockpit-update, qw(
    httpd reload
));

```

root/etc/e-smith/templates/etc/cockpit/cockpit.conf/10base
```diff
@@ -0,0 +1,60 @@
...
+
+        # skip if ipaddr is empty
+        my $ipaddr = $_->prop('ipaddr') || next;
+
+        push(@ips, $ipaddr);
+    }
+
+    # print results
+    foreach my $alias (@aliases) {
+        $OUT .= "https://$alias https://$alias:9090";
```
DavidePrincipi: Add the separator at the end to concatenate further for-cycles

```suggestion
        $OUT .= "https://$alias https://$alias:9090 ";
```

root/etc/e-smith/templates/etc/cockpit/cockpit.conf/10base
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    $OUT .= " ";
```
DavidePrincipi: No need to append it here.

```suggestion
```

root/etc/e-smith/templates/etc/cockpit/cockpit.conf/10base
```diff
@@ -0,0 +1,60 @@
...
+
+    foreach my $ip (@ips) {
+        $OUT .= "https://$ip https://$ip:9090";
```
DavidePrincipi: Add the separator at the end to concatenate further for-cycles

```suggestion
        $OUT .= "https://$ip https://$ip:9090 ";
```

root/etc/e-smith/templates/etc/httpd/conf.d/default-virtualhost.inc/15user_settings
```diff
@@ -0,0 +1,21 @@
...
+#
+
+ProxyPreserveHost On
+ProxyRequests Off
+
+RewriteEngine On
+RewriteCond %{HTTP:Upgrade} =websocket [NC]
+RewriteRule /(.*)           ws://127.0.0.1:9090/$1 [P,L]
+RewriteCond %{HTTP:Upgrade} !=websocket [NC]
+RewriteRule /(.*)           http://127.0.0.1:9090/$1 [P,L]
```
DavidePrincipi: I don't get why this block is needed, as you use the Proxy directives in the next lines.

I'm afraid mixing Rewrite* and Proxy* directives leads to configuration mistakes because they are processed with different priorities by the mod_rewrite and mod_proxy components. Can we use just _one_ approach?

I'd prefer the proxy/core one, if possible

gsanchietti: This is suggested by upstream: https://github.com/cockpit-project/cockpit/wiki/Proxying-Cockpit-over-Apache-with-LetsEncrypt

I'd leave it there, just to be sure :)

DavidePrincipi: Sorry I cannot find the lines below in the upstream suggestion. It's not the same code.

gsanchietti: It is the same code, at least diff command tells so.
I agree the rewrite rule is too wide, since the upstream code refers to a dedicated virtual host for cockpit.

edospadoni: I've tryed without the `RewriteRules`, it seems it works. We can remove the rewrites and test the redirect without them

root/etc/e-smith/templates/etc/httpd/conf.d/default-virtualhost.inc/15user_settings
```diff
@@ -0,0 +1,55 @@
...
+        if ((${'cockpit.socket'}{'UserSettingsGrantAccess'} || 'disabled') eq 'enabled') {
+
+            # check if external ips are set
+            if ((${'cockpit.socket'}{'UserSettingsGrantIPs'} || '') ne '') {
+                # split ips
+                my @ips = split(',', ${'cockpit.socket'}{'UserSettingsGrantIPs'});
+
+                # create require ip directive
+                foreach my $ip (@ips) {
+                    $OUT .= "    Require ip $ip\n";
```
DavidePrincipi: Does it make sense to (conditionally) add green and trusted networks too?

edospadoni: For the `Server Manager` access part on port `9090` we don't include Trusted Networks. I think we can use the same behaviour here

gsanchietti: We should see this page as a separate web application, I agree with @DavidePrincipi to support the trusted networks.
My proposal is:
- remove UserSettingsGrantIPs prop
- use local networks (see [roundcube](https://github.com/NethServer/nethserver-roundcubemail/tree/master/root/etc/e-smith/templates/etc/httpd/conf.d/roundcubemail.conf) as example)

edospadoni: Ok so if we enable external access, the only ips granted are the ips in the trusted networks list.

edospadoni:
> Also, please take care of new cockpit access policy: https://github.com/NethServer/nethserver-cockpit/blob/master/root/etc/e-smith/templates/etc/pam.d/cockpit/50Pam_listfile
> 
> How should we handle such case?

In this case the user can't login to cockpit if it isn't to a group specified in `/etc/nethserver/cockpit.allow` so the page is not accessible. Am i wrong?

DavidePrincipi:
> How should we handle such case?

We must add a new condition that **disables** that fragment if ``UserSettingsPage == 'enabled'``.

gsanchietti:
>  Am i wrong?

You are right.

> We must add a new condition that **disables** that fragment if `UserSettingsPage == 'enabled'`.

It makes sense, but should the configuration be reflected also on SSH page?

ui/public/i18n/language.json
```diff
@@ -1202,11 +1202,11 @@
     "AllowGroupsToSSH_SFTP":"Select one or more groups to grant SSH and/or SFTP access. Administrators are always granted SSH and SFTP access. The root account is still subject to the \"Allow root login\" checkbox setting.",
     "ssh_everyone_info": "Set the basic permission for non administrative accounts.",
     "ssh_administrators_info": "Administrators are always granted SSH and SFTP access.",
-    "shell_bash_needed": "The shell Bash is needed to use the cockpit server manager and the SSH service, you can override the shell of all users if this option is enabled",
+    "shell_bash_needed": "The shell Bash is needed to use the Cockpit Server Manager and the SSH service, you can override the shell of all users if this option is enabled. Once enabled, the Shell setting returned by the account provider is ignored. As a consequence, non-admin users must be granted again SSH and SFTP access permissions.",
```
DavidePrincipi: I think describing what happens when enabled is enough. I'd avoid the "you" form, too. The Bash requirement to access Cockpit is an internal detail we can avoid to communicate.
 
```suggestion
    "shell_bash_needed": "Once enabled, the Shell setting returned by the account provider is ignored. As a consequence, non-admin users must be granted again SSH and SFTP access permissions.",
```

What do you think?

edospadoni: I agree :+1:

PR #241, opened 2020-03-31
UI & API: handle restore with --skip-network

gsanchietti:
Preserve existing restore flow, add an option to skip network restore.

NethServer/dev#6099

First step, the "restore network" is check by default:
Screenshot from 2020-03-31 15-37-11

After un-checking the checkbox:
Screenshot from 2020-03-31 15-37-22


ui/public/i18n/language.json
```diff
@@ -319,7 +319,9 @@
     "invalid_backup_format": "Backup not found or invalid",
     "remap_at_least_one_green": "You have to select at least one green to remap",
     "assign_later": "Assign later",
-    "current_int": "System interfaces"
+    "current_int": "System interfaces",
+    "restore_network": "Restore network configuration",
+    "skip_network_info": "Current network configuration will per preserved, but some applications may be disabled at the end of the process. After restore, review the system configuration to re-enable all required applications."
```
filippocarletti:
```suggestion
    "skip_network_info": "Current network configuration will be preserved, but some applications may be disabled at the end of the process. After restore, review the system configuration to re-enable all required applications."
```

ui/public/i18n/language.json
```diff
@@ -319,7 +319,9 @@
     "invalid_backup_format": "Backup not found or invalid",
     "remap_at_least_one_green": "You have to select at least one green to remap",
     "assign_later": "Assign later",
-    "current_int": "System interfaces"
+    "current_int": "System interfaces",
+    "restore_network": "Restore network configuration",
+    "skip_network_info": "Current network configuration will be preserved, but some applications may be disabled at the end of the process. After restore, review the system configuration to re-enable all required applications."
```
DavidePrincipi: Is it better with gerund?

```suggestion
    "skip_network_info": "Current network configuration will be preserved, but some applications may be disabled at the end of the process. After restoring, review the system configuration to re-enable all required applications."
```

gsanchietti:
> I think a switch is needed instead of a checkbox because it enables/disables another form section.

Switches are used for progressive disclosure, and this is not the case: Is's a change of behavior of the later restore process.

Also the checkbox is more uniform with the entire dialog style.

edospadoni:
> > I think a switch is needed instead of a checkbox because it enables/disables another form section.
> 
> Switches are used for progressive disclosure, and this is not the case: Is's a change of behavior of the later restore process.
> 
> Also the checkbox is more uniform with the entire dialog style.

I agree, a switch is too much invasive in this case.

PR #250, opened 2020-04-09
Custom error for wrong old password

andre8244:
Show the user a specific error when he/she tries to change the password but has entered a wrong old password.

https://github.com/NethServer/dev/issues/6125


edospadoni:
Ok for me, but i'm waiting @gsanchietti for [this](https://github.com/NethServer/nethserver-cockpit/pull/250/files#diff-fe304f5d5ebf1390a2eddf9aa2729865R64) part

PR #252, opened 2020-04-24
perl helper: UTF8 fixes

gsanchietti:
- make sure STDIN is read using UTF8
- add printJson function: use it when printing UTF8 encoded strings

NethServer/dev#6128

### DIFF
--- a/lib/perl/NethServer/ApiTools.pm
+++ b/lib/perl/NethServer/ApiTools.pm
@@ -21,7 +21,7 @@
 use strict;
 package NethServer::ApiTools;
 use Exporter 'import';
-our @EXPORT_OK = qw(hints success error readInput safe_decode_json invoke_info_api);
+our @EXPORT_OK = qw(hints success error readInput safe_decode_json invoke_info_api printJson);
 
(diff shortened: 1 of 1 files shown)


lib/perl/NethServer/ApiTools.pm
```diff
@@ -99,7 +99,12 @@ sub error
 #
 sub readInput
 {
-    my $str = do { local $/; <STDIN> };
+
+    my $tmp = do { local $/; <STDIN> };
+
+    # Make sure everything is UTF-8 encoded
+    use Encode qw(encode);
+    my $str = encode("UTF-8", $tmp);
```
DavidePrincipi, line 107: what happens if $tmp has a wrong encoding? does Perl die()?

gsanchietti: No it doesn't die, but we can use CROAK if we want to: https://perldoc.perl.org/Encode.html#Handling-Malformed-Data

gsanchietti:
Closing for now, see comment inside the linked issue.

PR #255, opened 2020-05-05
ui. remove useless change hostname message

edospadoni:
Cockpit is no more restarted after hostname change.


gsanchietti:
Released:
nethserver-cockpit-1.6.5-1.ns7.noarch.rpm

PR #261, opened 2020-05-16
ui. fix storage info view

federicoballarini:
Fixed storage info bar view issue
Refer to https://github.com/NethServer/dev/issues/6168

### DIFF
--- a/ui/src/components/system/Backup.vue
+++ b/ui/src/components/system/Backup.vue
@@ -2908,6 +2908,7 @@ export default {
               : false;
 
           // visualize storage info
+          $.fn.popover.Constructor.DEFAULTS.whiteList["*"].push('style');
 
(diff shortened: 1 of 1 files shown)


federicoballarini:
I think @gsanchietti and @edospadoni can review.

gsanchietti:
Sorry, but I really do not know what the patch does.
Let's see if someone else with more skill can help here :)

federicoballarini:
> Sorry, but I really do not know what the patch does.

This should allow to use inline style property into a popover. Otherwise it will be blocked. And this is the cause of the bug: the inline style is deleted if you check the code.

edospadoni:
> I think @gsanchietti and @edospadoni can review.

The patch is cool :+1: Great job @federicoballarini

PR #264, opened 2020-05-20
New policy 20200510 only TLS 1.2

stephdl:
NethServer/dev#6170

### DIFF
--- a/ui/public/i18n/language.json
+++ b/ui/public/i18n/language.json
@@ -783,10 +783,10 @@
     "default_policy": "Default upstream policy",
     "tls_policy_edit_ok": "TLS policy updated successfully!",
     "tls_policy_edit_error": "TLS policy not updated.",
-    "20200510": "20200510 only TLS 1.2",
-    "20180621": "20180621",
-    "20180330": "20180330",
-    "20181001": "20181001"
+    "20200510": "2020-05-10 only TLS 1.2",
+    "20180621": "2018-06-21",
+    "20180330": "2018-03-30",
+    "20181001": "2018-10-01"
   },
   "trusted_networks": {
     "title": "Trusted networks",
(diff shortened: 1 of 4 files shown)


ui/public/i18n/language.json
```diff
@@ -782,7 +782,11 @@
     "enforce_security": "Enforced security level",
     "default_policy": "Default upstream policy",
     "tls_policy_edit_ok": "TLS policy updated successfully!",
-    "tls_policy_edit_error": "TLS policy not updated."
+    "tls_policy_edit_error": "TLS policy not updated.",
+    "20200510": "20200510 only TLS 1.2",
+    "20180621": "20180621",
+    "20180330": "20180330",
+    "20181001": "20181001"
```
DavidePrincipi, line 789:
```suggestion
    "20200510": "2020-05-10 only TLS 1.2",
    "20180621": "2018-06-21",
    "20180330": "2018-03-30",
    "20181001": "2018-10-01"
```

ui/src/components/system/TLSPolicy.vue
```diff
@@ -149,6 +149,9 @@ export default {
           context.TLSPolicy.policies = context.TLSPolicy.policies.concat(
             success.status.available
           );
+          // Sort the array by numerical ascending order
+          context.TLSPolicy.policies.sort((a,b) => a-b);
+
```
DavidePrincipi, line 154: I don't get why we need to sort the array. The items order is already hardcoded here: https://github.com/NethServer/nethserver-cockpit/pull/264/files#diff-0993a3b6f24ef6461931c72ef20d0418R31

I'd prefer to fix it (if necessary) in that API `read` script.

```suggestion
```

stephdl: when I tested my code I saw that the array is not in the good order, I can let is as is or try to fix it. I suppose that javascript and bash did not agree on the type of variable. My fix is probably not the best but we have to fix something.

stephdl: yep it should be fixed, I put my glasses on my nose

PR #265, opened 2020-05-25
api & ui. show all logs of a specific backup

edospadoni:
Nethserver/dev#6175


api/system-backup/read
```diff
@@ -104,6 +104,28 @@ sub find_latest_log
     return ($latest_filename, $latest_ts);
 }
 
+sub find_all_logs
+{
+    my $name = shift;
+    my $n = "[0123456789]";
```
filippocarletti, line 110:
```suggestion
    my $n = "[0-9]";
```

api/system-backup/read
```diff
@@ -104,6 +104,28 @@ sub find_latest_log
...
+    my $logs;
+    foreach my $file (glob("/var/log/backup/backup-$name-$n$n$n$n$n$n$n$n$n$n$n$n.log")) {
```
filippocarletti, line 112:
```suggestion
    foreach my $file (glob("/var/log/backup/backup-$name-$n.log")) {
```
While exact digits count may be more accurate, I think it's safe enough getting all digits.
Removing $n (remove also line 110) should improve clarity.

edospadoni: I'm not sure this change works :thinking: because it's not a regex, let me try. However I copied the previous method which did the same :grin:

gsanchietti: Proposed match will not work, you need to use:
```
/var/log/backup/backup-$name-$n*.log
```
Still, I prefer the current implementation: it will match exactly needed files.

api/system-backup/read
gsanchietti: I agree on this.

api/system-backup/read
```diff
@@ -104,6 +104,28 @@ sub find_latest_log
     return ($latest_filename, $latest_ts);
 }
```
gsanchietti, line 105: I think we can totally remove this code along with `last-log` action: it seems the UI doesn't use it anymore.
If the UI still use it, please merge `find_all_logs and `find_latest_log` in one function where `find_all_logs` return an ordered array.

api/system-backup/read
```diff
@@ -104,6 +104,28 @@ sub find_latest_log
...
+    my $n = "[0123456789]";
+    my $logs;
+    foreach my $file (glob("/var/log/backup/backup-$name-$n$n$n$n$n$n$n$n$n$n$n$n.log")) {
+        my $stat = stat($file);
+        if ( -f $file ) {
+            open my $fh, '<', $file;
+            $/ = undef;
+
+            $file =~ s/[^0-9]//g;
+            my ($y,$M,$d,$h,$m) = $file =~ /^([0-9]{4})([0-9]{2})([0-9]{2})([0-9]{2})([0-9]{2})\z/ or die;
```
gsanchietti, line 119: Never die inside an API or it will return a malformed JSON for the UI.
In case of error, use the `error()` function.

api/system-backup/read
edospadoni: I think it's a good pattern to never remove an API, because someone can use it in an external way (scripts, etc...). UI doesn't need anymore the API but I would like to keep it. However, we can merge it in one function to clean the code

api/system-backup/read
```diff
@@ -88,44 +88,51 @@ sub get_default_custom_excludes
...
+
+            push(@logs, {"date" => "$y-$M-$d $h:$m", "data" => <$fh>, "path" => $path, "ts" => $stat->mtime});
+
+            close $fh;
         }
     }
-    return ($latest_filename, $latest_ts);
+
+    my @sorted = reverse sort { $a->{ts} <=> $b->{ts} } @logs;
+    return @sorted;
```
gsanchietti, line 113: This should do the same:
```suggestion
    return reverse sort { $a->{ts} <=> $b->{ts} } @logs;
```

api/system-backup/read
```diff
@@ -88,44 +88,51 @@ sub get_default_custom_excludes
...
-            $latest_filename = $file
+        if ( -f $file ) {
+            open my $fh, '<', $file;
+            $/ = undef;
+
+            my $path = $file;
+            $file =~ s/[^0-9]//g;
+            my ($y,$M,$d,$h,$m) = $file =~ /^([0-9]{4})([0-9]{2})([0-9]{2})([0-9]{2})([0-9]{2})\z/ or error();
+
+            push(@logs, {"date" => "$y-$M-$d $h:$m", "data" => <$fh>, "path" => $path, "ts" => $stat->mtime});
```
gsanchietti, line 106: Do not slurp file: this API will soon fill the system machine RAM if there are many backup files (eg. a backup scheduled once per hour).

edospadoni: To avoid this, we can call asynchronously from the UI the content of the log when the user change the backup date selector.

gsanchietti: I like the idea!

PR #266, opened 2020-05-25
New policy 20200510 only tls1.2

stephdl:
https://github.com/NethServer/dev/issues/6170

### DIFF
--- a/createlinks
+++ b/createlinks
@@ -43,6 +43,7 @@ event_templates('nethserver-cockpit-update', qw(
     /etc/httpd/conf.d/default-virtualhost.inc
     /etc/pam.d/cockpit
     /etc/nethserver/cockpit.otp.force
+    /etc/sysconfig/cockpit
 ));
 
 event_services('nethserver-cockpit-update', qw(
@@ -168,3 +169,11 @@ safe_symlink('/bin/true', 'root/etc/e-smith/events/test-success/S20true');
 safe_symlink('/bin/false', 'root/etc/e-smith/events/test-failure/S20false');
 safe_symlink('/bin/true', 'root/etc/e-smith/validators/test-success/S20true');
 safe_symlink('/bin/false', 'root/etc/e-smith/validators/test-failure/S20false');
+
+#
+# tls-policy-save event
+#
+
+event_templates('tls-policy-save', qw(
+    /etc/sysconfig/cockpit
+));
--- a/root/etc/systemd/system/cockpit.service.d/nethserver.conf
+++ b/root/etc/systemd/system/cockpit.service.d/nethserver.conf
@@ -1,0 +1,2 @@
+[Service]
+EnvironmentFile=-/etc/sysconfig/cockpit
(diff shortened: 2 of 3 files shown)


createlinks
```diff
@@ -168,3 +168,15 @@ safe_symlink('/bin/true', 'root/etc/e-smith/events/test-success/S20true');
...
+#
+# tls-policy-save event
+#
+
+event_templates('tls-policy-save', qw(
+        /etc/systemd/system/cockpit.service.d/nethserver.conf
+));
+
+event_services('tls-policy-save', qw(
+        cockpit restart
```
gsanchietti, line 181: Do not restart Cockpit, otherwise the `tls-policy-save` will be interrupted if called from Cockpit itself.
Cockpit will apply the changes on next restart, 90seconds after the last users has been logged off.

stephdl: how to say to systemctl that the daemon need to be reloaded by `systemctl daemon reload`. It is included by the perl api of nethserver 

It is necessary to add optional settings to systemd

gsanchietti: The `daemon-reload` is already handled by the runlevel-adjust (or equivalent action).

DavidePrincipi: I agree with Giacomo. Do not restart cockpit: it is not a long-running daemon, it stops automatically when the session ends and it is idle for 90 seconds.

stephdl: yep, giacomo show me another unknow way for me with a Env file, I think it will be better

stephdl: The env file is created as a template

stephdl:
When the template of the ENV file is created the service is not restarted but it seems when the root login (I suppose if no other users are connected) then the service is restarted automatically so the ENV file is loaded to the service

root/etc/systemd/system/cockpit.service.d/nethserver.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+EnvironmentFile=-/etc/sysconfig/cockpit
```
DavidePrincipi, line 2: This shouldn't be effective until next `systemctl daemon-reload`. However I wouldn't fix it because it is required just once, and it surely occurs when one of the following things happens:

- system reboot
- an RPM containing a systemd unit is updated

As the current Cockpit policy is far from being considered unsecure I'd not worry...

```suggestion
EnvironmentFile=-/etc/sysconfig/cockpit
```

gsanchietti:
Wait release of 6175 before merge.

root/etc/systemd/system/cockpit.service.d/nethserver.conf
stephdl: just trying to reproduce with a fresh install @DavidePrincipi 

install the rpm of this PR
test the cipher, tls 1.1 and 1.2 are allowed
move to the policy 20200510
wait (a long time) the service restart alone, about 5 minutes. You can try to login root several time
test the cipher suite, we are only allowed to tls1.2

It seems we have nothing more to do indeed

gsanchietti:
Ready for merge. It's all yours @stephdl ;)

PR #269, opened 2020-05-28
Networks page: Properly show hotspot interface

andre8244:
- Display hotspot interface with a proper icon and a distinctive color
- ~Show `tun-dedalo` CIDR address~

https://github.com/NethServer/dev/issues/6186

### DIFF
--- a/api/system-network/read
+++ b/api/system-network/read
@@ -160,12 +160,7 @@ if($cmd eq 'list') {
         }
 
         # read address in cidr format
-        my $iname_for_cidr = $iname;
-
-        if ($props{'role'} eq 'hotspot') {
-            $iname_for_cidr = 'tun-dedalo';
-        }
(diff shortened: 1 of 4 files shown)


andre8244:
> I have mixed feelings on this one.
> While the patch for sure reaches the goal, you're hard-coding stuff from an extra package (nethserver-dedalo, not widely used) straight into the core.
> 
> IMO this is not acceptable from an architecture point of view

You are right, I didn't think about this architectural aspect. Is your concern about the tun-dedalo CIDR retrieval only, or to the **hotspot** role management as a whole? Not showing hotspot CIDR could be a good compromise?

gsanchietti:
> You are right, I didn't think about this architectural aspect. Is your concern about the tun-dedalo CIDR retrieval only, or to the **hotspot** role management as a whole? Not showing hotspot CIDR could be a good compromise?

The hotspot role as a whole, I didn't never liked it. Still, it's there since NethServer 6 and the role didn't change the name so I think it's stable enough.
Let's go with displaying just not displaying the CIDR, so you do not need to hard-code `tun-dedalo` interface name ;)

PR #275, opened 2020-06-30
Support for shared i18n strings

andre8244:
Every Cockpit module has access to a set of shared i18n strings provided by `ui/public/i18n/common.json` file.

https://github.com/NethServer/dev/issues/6206


andre8244:
In the end we decided to discard support for shared i18n strings in order to avoid unnecessary complexity

PR #283, opened 2020-07-17
language. added missing label

federicoballarini:
Added missing "users_groups.ShellOverride" label

### DIFF
--- a/ui/public/i18n/language.json
+++ b/ui/public/i18n/language.json
@@ -641,7 +641,8 @@
     "UserDN": "User DN",
     "GroupDN": "Group DN",
     "StartTls": "Start TLS",
-    "edit_delegation": "Edit delegations of {name}"
+    "edit_delegation": "Edit delegations of {name}",
+    "ShellOverride": "Shell override"
   },
   "network": {
     "title": "Network",


edospadoni:
Sorry but I can't find the view in where the label is missing, can you provide a screenshot or a view reference? Thanks

federicoballarini:
> Sorry but I can't find the view in where the label is missing, can you provide a screenshot or a view reference? Thanks

Sure! :-)

image

You can find it under System > Users and groups > Provider details

edospadoni:
Thanks @federicoballarini :+1:

gsanchietti:
Sorry to be late at the party, but IMO the patch is not good.
Indeed it adds the label but that property should not be displayed there.

So the real fix should be:
- update the UI (or API) to not display shell override
- remove the new label

edospadoni:
So like other params excluded here: https://github.com/NethServer/nethserver-cockpit/blob/master/ui/src/components/system/UsersGroups.vue#L57 we have to exclude also `ShellOverride`. Only UI fix is necessary

gsanchietti:
> So like other params excluded here: https://github.com/NethServer/nethserver-cockpit/blob/master/ui/src/components/system/UsersGroups.vue#L57 we have to exclude also `ShellOverride`. Only UI fix is necessary.

Exactly!

federicoballarini:
If you want I open a new PR. Tell me only a thing... I have to remove shellOverride parameter from line 57 and also from line 60, correct?

gsanchietti:
> If you want I open a new PR. Tell me only a thing... I have to remove shellOverride parameter from line 57 and also from line 60, correct?

I do not know. @andre8244 @edospadoni any help? :)

edospadoni:
> If you want I open a new PR. Tell me only a thing... I have to remove shellOverride parameter from line 57 and also from line 60, correct?

You have to hide the `ShellOverride` param from the UI, using the code linked above. The API returns many parameters, such as `NsdcIp`,` ValidHostname` or `AuthRequired` and these parameters are not displayed in the UI because the condition on Vue hides them.

You can open a new PR with a patch like this:
```js
v-if="k != 'oldIp' && ... && k != 'ShellOverride'"
```

PR #284, opened 2020-07-20
ui. remove ShellOverride from provider details

federicoballarini:
Removed ShellOverride from provider details and unused language string.
Continuing from https://github.com/NethServer/nethserver-cockpit/pull/283


edospadoni:
Before 
before

After
after

Good job :+1:

PR #286, opened 2020-08-05
Display the base32 secret

stephdl:
Display the base32 key, it can be used to be stored it manually

Screenshot (46)

https://community.nethserver.org/t/2fa-or-two-factor-authentication-with-cockpit/14172/109?u=stephdl


stephdl:
We need more review of the support team, closed by now, waiting more informations

PR #296, opened 2020-10-29
services. fix delegation on access field

gsanchietti:
Before this change, the delegated user could not access the list of
available roles and zones.

NethServer/dev#6316

### DIFF
--- a/api/system-services/read
+++ b/api/system-services/read
@@ -26,6 +26,7 @@ use strict;
 use warnings;
 
 use esmith::ConfigDB;
+use esmith::NetworksDB;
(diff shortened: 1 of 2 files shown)


gsanchietti:
While this fix works, I think we need a better solution. See comment on the issue.

gsanchietti:
I can't see any better solution on this, let's go to review.

edospadoni (approved):

edospadoni:
I think this is the only possibile solution if the `nethserver-firewall-base-ui` package is not installed.

andre8244 (approved):

PR #297, opened 2020-11-12
Fix ShellOverride regression

DavidePrincipi:
After ShellOverride was introduced by https://github.com/NethServer/nethserver-cockpit/pull/209 the remote LDAP account provider procedure shows a "Shell override" field, as reported by https://github.com/NethServer/dev/issues/6327

This PR fixes the JSON object content returned by the API to avoid that regression.

Furthermore, it fixes the form labels and hides the LdapUriDn field, which is not required by the bind procedure: see indidivual commits for details.

This is a screenshot of the results (compare it with the issue attachment).

image

NethServer/dev#6327

### DIFF
--- a/ui/src/components/system/UsersGroups.vue
+++ b/ui/src/components/system/UsersGroups.vue
@@ -1535,7 +1535,7 @@
                           </div>
                         </div>-->
                         <div
-                          v-if="newProvider.isChecked && k!='action' && k != 'NsdcIp' && k!='DiscoverDcType' && k!='port' && k!='IsLocal' && k!='isAD' && k!='isLdap' && k!='host' && k!='Provider' && !((newProvider.info.BindType == 'anonymous' && k=='BindPassword') || (newProvider.info.BindType == 'anonymous' && k=='BindDN'))"
+                          v-if="newProvider.isChecked && k!='action' && k != 'LdapUriDn' && k != 'NsdcIp' && k!='DiscoverDcType' && k!='port' && k!='IsLocal' && k!='isAD' && k!='isLdap' && k!='host' && k!='Provider' && !((newProvider.info.BindType == 'anonymous' && k=='BindPassword') || (newProvider.info.BindType == 'anonymous' && k=='BindDN'))"
                           v-for="(v,k) in newProvider.info"
                           v-bind:key="k"
                           class="form-group"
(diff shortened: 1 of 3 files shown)


ui/src/components/system/UsersGroups.vue
```diff
@@ -1535,15 +1535,15 @@
...
-                          v-if="newProvider.isChecked && k!='action' && k != 'NsdcIp' && k!='DiscoverDcType' && k!='port' && k!='IsLocal' && k!='isAD' && k!='isLdap' && k!='host' && k!='Provider' && !((newProvider.info.BindType == 'anonymous' && k=='BindPassword') || (newProvider.info.BindType == 'anonymous' && k=='BindDN'))"
+                          v-if="newProvider.isChecked && k!='action' && k != 'LdapUriDn' && k != 'NsdcIp' && k!='DiscoverDcType' && k!='port' && k!='IsLocal' && k!='isAD' && k!='isLdap' && k!='host' && k!='Provider' && !((newProvider.info.BindType == 'anonymous' && k=='BindPassword') || (newProvider.info.BindType == 'anonymous' && k=='BindDN'))"
                           v-for="(v,k) in newProvider.info"
                           v-bind:key="k"
                           class="form-group"
                         >
                           <label
                             class="col-sm-3 control-label"
                             for="textInput-modal-markup"
-                          >{{k | camelToSentence}}</label>
```
gsanchietti, line 1546: I can't understand what is the effect of this change. 
Could you please provide and emaple?

gsanchietti (approved):

edospadoni (approved):

edospadoni:
PR is ok, but using `camelToSentence` filter we remove [this](https://github.com/NethServer/nethserver-cockpit/blob/83e92ff587690cea3b5e220d93b3b2a343fe1be4/ui/public/i18n/language.json#L640) label. Because `NsdcIp ` with `camelToSentence` is "Nsdc Ip"  and not "Active Directory IP". It's right?

ui/src/components/system/UsersGroups.vue
DavidePrincipi: I added a screenshot to the PR description

DavidePrincipi: I added a screenshot to the PR description

DavidePrincipi:
That label is used here

image

It's still in place.

I checked the other account provider configurations and `NsdcIp | camelToSentence` is never used.

PR #298, opened 2020-11-18
Fix undefined references

DavidePrincipi:
```
[root@vm6 ~]# echo '{...}' | /usr/bin/sudo /usr/libexec/nethserver/api/system-backup/validate | jq
PHP Warning:  fsockopen(): unable to connect to unix:///var/run/smwingsd.sock:-1 (No such file or directory) in /usr/libexec/nethserver/api/lib/EsmithDatabase.php on line 222
PHP Fatal error:  Call to undefined method EsmithDatabase::getLog() in /usr/libexec/nethserver/api/lib/EsmithDatabase.php on line 224
```

Since smwingsd is not available by default, we must remove log calls as they are not defined in this context.

Regression of https://github.com/NethServer/dev/issues/6320

### DIFF
--- a/api/lib/EsmithDatabase.php
+++ b/api/lib/EsmithDatabase.php
@@ -219,7 +219,7 @@ class EsmithDatabase {
             $socketPath = '/var/run/smwingsd.sock';
             $errno = 0;
             $errstr = '';
-            self::$socket = fsockopen('unix://' . $socketPath, -1, $errno, $errstr);
+            self::$socket = @fsockopen('unix://' . $socketPath, -1, $errno, $errstr);
         }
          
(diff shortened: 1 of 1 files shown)


gsanchietti (commented):
If `nethserver-httpd-admin` is not installed, all PHP validators will always rise this warning on stderr:
```
PHP Warning:  fsockopen(): unable to connect to unix:///var/run/smwingsd.sock:-1 (Connection refused) in /usr/libexec/nethserver/api/lib/EsmithDatabase.php on line 222
```

Can we avoid it?

gsanchietti (approved):

PR #302, opened 2020-12-01
Use warning instead doc-info for restic secret

stephdl:
I removed the doc-info and display the warning in the modal, I cannot get it in work

Screenshot (22)

https://github.com/NethServer/dev/issues/6346


gsanchietti (approved):
Clever workaround ;)

PR #303, opened 2020-12-04
Delete a certificate with the UI

stephdl:
Currently, the admin can delete a certificate only using the commands documented here: https://docs.nethserver.org/en/v7/base_system.html#disable-let-s-encrypt

With this PR, the deletion could be applied to any non-used certificate.

https://github.com/NethServer/dev/issues/6362

### DIFF
--- a/api/system-certificate/update
+++ b/api/system-certificate/update
@@ -63,4 +63,17 @@ case $action in
         /sbin/e-smith/signal-event -j certificate-update
        ;;
 
+    "delete")
+
+        file=$(echo $data | jq -r '.file')
+        key=$(echo $data | jq -r '.key')
+        chain=$(echo $data | jq -r '.chain')
+
+        /usr/bin/rm -f "$file"
+        /usr/bin/rm -f "$key"
+        /usr/bin/rm -f "$chain"
+
+        ;;
+
+
 esac
--- a/ui/src/components/system/Certificates.vue
+++ b/ui/src/components/system/Certificates.vue
@@ -113,8 +113,8 @@
                   {{$t('certificates.set_as_default')}}
                 </a>
               </li>
-              <li v-if="!props.row.default && props.row.file !== '/etc/pki/tls/certs/NSRV.crt'" role="presentation" class="divider"></li>
-              <li v-if="!props.row.default && props.row.file !== '/etc/pki/tls/certs/NSRV.crt'">
+              <li v-if="!props.row.default && !props.row.builtin" role="presentation" class="divider"></li>
+              <li v-if="!props.row.default && !props.row.builtin">
                 <a @click="openDeleteModal(props.row)">
                   <span class="fa fa-times span-right-margin"></span>
                   {{$t('certificates.delete')}}
--- a/ui/src/components/system/Certificates.vue
+++ b/ui/src/components/system/Certificates.vue
@@ -113,6 +113,13 @@
                   {{$t('certificates.set_as_default')}}
                 </a>
               </li>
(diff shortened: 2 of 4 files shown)


api/system-certificate/update
```diff
@@ -63,4 +63,20 @@ case $action in
...
+        key=$(echo $data | jq -r '.key')
+        chain=$(echo $data | jq -r '.chain')
+
+        /usr/bin/rm -f "$file"
+        /usr/bin/rm -f "$key"
+        /usr/bin/rm -f "$chain"
+
+        if  [[ $file =~ '/etc/letsencrypt/' ]]; then
+            /usr/sbin/e-smith/config setprop pki LetsEncryptDomains ''
+        fi
```
gsanchietti, line 78: These lines has a serious side effect.

You can have this scenario:
- certificate `le1` with domains `first.domain.org`
- then you add new domain to to `le1` and it becomes `le2`: `first.domain.org, second.domain.org`
- set `le1` as default certificate
- delete `le1` from the web interface
- `le2` will not be renewed anymore

This is not clear even from the doc, but in this case you're not only deleting a certificate, you're also disabling LE at all.

So let's restrict this PR to only deleting the selected certificate.
@filippocarletti do you want to implement another button to totally disable LE?

ui/src/components/system/Certificates.vue
```diff
@@ -113,6 +113,13 @@
                   {{$t('certificates.set_as_default')}}
                 </a>
               </li>
+              <li v-if="!props.row.default && props.row.file !== '/etc/pki/tls/certs/NSRV.crt'" role="presentation" class="divider"></li>
+              <li v-if="!props.row.default && props.row.file !== '/etc/pki/tls/certs/NSRV.crt'">
```
gsanchietti, line 117: Can we avoid harcoding that path one the UI?
Let's move the info on the API.

Insie the `read`, you can loop over existing certificates and add a field to JSON named like `builtin` to indicate which is the non-deletable certificate.

gsanchietti (changes requested):

api/system-certificate/update
stephdl: > set le1 as default certificate
> delete le1 from the web interface

the certificate set as default cannot be deleted, does it is a typo @gsanchietti

gsanchietti: Yes, corrected.

Let's wait thoughts from @filippocarletti

ui/src/components/system/Certificates.vue
stephdl: the elegant way could be to use `/usr/libexec/nethserver/cert-list` from nethserver-base, what do you think ?

gsanchietti: We already use it and parse the output of the command inside `read`.
I guess a modification only inside the API is simpler, but you could try both ways

stephdl: Yes it is simpler, I implemented it

api/system-certificate/update
stephdl: this part is still waiting some inputs

filippocarletti: @gsanchietti is right, we can't set the LetsEncryptDomains pki prop empty, because the renew event exits if it's empty and the cert will not be renewed.

gsanchietti: Ok, so let's restrict the patch to only removing the cert without editing LE props.

Should the doc changed as well?

stephdl: change done as requested

filippocarletti (approved):

gsanchietti (approved):

PR #305, opened 2020-12-15
user-settings: disable ProxyPreserveHost

gsanchietti:
If ProxyPreserveHost is enabled, it will change the behavior of all
proxy pass configured after this fragment.

This option may change the behavior of backend HTTP servers,
like Tomcat which respond with a wrong Location header

NethServer/dev#6364


DavidePrincipi (commented):
Is there an alternative implementation, by the side of Apache configuration for WebTop?

DavidePrincipi (approved):

PR #323, opened 2021-02-27
ui. show total of users and groups

federicoballarini:
Print the total of users and groups in Cockpit interface.
Refer to https://github.com/NethServer/dev/issues/6443

### DIFF
--- a/ui/public/i18n/language.json
+++ b/ui/public/i18n/language.json
@@ -667,7 +667,8 @@
     "GroupDN": "Group DN",
     "StartTls": "STARTTLS",
     "BindType": "Bind type",
-    "edit_delegation": "Edit delegations of {name}"
+    "edit_delegation": "Edit delegations of {name}",
+    "total": "Total"
   },
   "network": {
     "title": "Network",
--- a/ui/src/components/system/UsersGroups.vue
+++ b/ui/src/components/system/UsersGroups.vue
@@ -164,6 +164,9 @@
           </li>
         </ul>
       </div>
+      <span class="users-count">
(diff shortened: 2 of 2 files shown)


edospadoni:
Following Patternfly's guidelines, it's better to add the number of a particular list next to the search box like here:
Screenshot from 2021-03-01 09-08-35

In the specific case of the Users and Groups page, the search box acts dynamically based on the type of filter:
Users:
Screenshot from 2021-03-01 09-10-42
Groups:
Screenshot from 2021-03-01 09-10-54

So we can add the number of users or groups next to the search box once for Users (in case "Users" type is selected) once for Groups (in case "Groups" type is selected).

federicoballarini:
@edospadoni what do you think now? :)

edospadoni:
> @edospadoni what do you think now? :)

Much better now. I prefer to change Users: N and Groups: N to Total: N (for both users and groups) because it looks like a repetition. What do you think?

edospadoni:
> > > @edospadoni what do you think now? :)
> > 
> > 
> > Much better now. I prefer to change Users: N and Groups: N to Total: N (for both users and groups) because it looks like a repetition. What do you think?
> 
> Mmm... I don't know... For clarity it's better to show "Users" and "Groups" in my opinion, but I leave the choice to you :)

In other pages or Cockpit applications we have always used "Total". Let's go with Total also here.

federicoballarini:
Here we are @edospadoni :)

edospadoni (approved):

edospadoni:
Good job :+1:

PR #325, opened 2021-03-09
Do not block to disable smarthosts

stephdl:
The save button was in a disabled state until the smarthost settings are not validated, now we take in consideration that the SmartHostStatus must be true also 

https://github.com/NethServer/dev/issues/6450


gsanchietti (approved):

gsanchietti (commented):
Mmm wait, I need to check something

gsanchietti:
Ok, the logic seems good.

PR #326, opened 2021-04-13
Local Rules: Force to write access property

stephdl:
When you modify a  Local rules inside the firewall UI, we force to delete the access property when we allow to connect only from localhost

[In the validation](https://github.com/NethServer/nethserver-cockpit/blob/master/api/system-services/validate#L127) we check that the property must exist in the json string, so it is safe to write anyway the access prop empty

https://github.com/NethServer/dev/issues/6482

### DIFF
--- a/api/system-services/update
+++ b/api/system-services/update
@@ -77,13 +77,8 @@ case $action in
             fi
         fi
 
-        if [ -n "$access" ]; then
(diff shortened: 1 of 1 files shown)


gsanchietti (approved):
:+1: you can open the issue

PR #327, opened 2021-04-14
css: extend left menu nav bar

gsanchietti:
Try to avoid truncated labels for languages like Spanish

See https://community.nethserver.org/t/fix-correct-spanish-translation/18056/14?u=giacomo


gsanchietti:
Replaced by #331

PR #333, opened 2021-04-16
Update brandings.css

jgjimenez:
Login page's labels get cut off in certain languages

### DIFF
--- a/root/usr/share/cockpit/branding/nethserver/branding.css
+++ b/root/usr/share/cockpit/branding/nethserver/branding.css
@@ -220,3 +220,23 @@ input[type="checkbox"].form-control {
         grid-row: 3/4 !important;
     }
 }
+
+@media (min-width: 768px){
+.login-pf .container .login-area {
+    border-right: 1px solid #474747;
+    position: relative;
+    min-height: 1px;
+    padding-left: 0px;
+    padding-right: 14px;
+}
+}
+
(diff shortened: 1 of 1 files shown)


root/usr/share/cockpit/branding/nethserver/branding.css
```diff
@@ -220,3 +220,23 @@ input[type="checkbox"].form-control {
...
+}
+
+input[type="text"].form-control { 
+width: 90%;
+margin-left: auto;
+}
+input[type="password"].form-control {
+width: 90%;
+margin-left: auto;
+}
```
gsanchietti, line 241: For sure these selector are way too wide: you're going to change almost every form in the UI.
You should narrow the scope of the selectors.

gsanchietti (changes requested):

root/usr/share/cockpit/branding/nethserver/branding.css
jgjimenez: I am going to leave it without effect, what is happening is not css problems, by increasing the resolution of my screen it is possible to see well.

jgjimenez: it is not necessary to apply it for production

gsanchietti:
Closed as per request of @jgjimenez

PR #341, opened 2021-07-01
nethserver-httpd must not be removed even by root

stephdl:
nethserver-httpd is an application that it should be never removed even by root.

Some others application like Let'sencrypt could rely on it

https://github.com/NethServer/dev/issues/6540


andre8244 (approved):

gsanchietti (approved):
Please open the issue for it :)

edospadoni (approved):

PR #347, opened 2021-11-08
Validate settings of smarthost with swaks

stephdl:
swaks just try to authenticate with the credential and exit with 0 or not 0 code. It is better than send an email that we could not assume it could be delivered.  

https://github.com/NethServer/dev/issues/6594

### DIFF
--- a/api/system-settings/execute
+++ b/api/system-settings/execute
@@ -42,11 +42,11 @@ case $action in
         echo -e $text | /usr/lib/sendmail -F $hostname -f $from $email 2>&1
         ;;
     "test-smarthost")
-        hostname=$(echo $data | jq -r '.SmartHostName')
-        username=$(echo $data | jq -r '.SmartHostUsername')
-        password=$(echo $data | jq -r '.SmartHostPassword')
-        port=$(echo $data | jq -r '.SmartHostPort')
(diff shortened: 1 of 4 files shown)


api/system-settings/execute
```diff
@@ -47,15 +47,6 @@ case $action in
...
-Date: $(date -R)
-Message-ID: testEmail.$(date +%s)@$(hostname -d)
-From: no-reply@$(hostname -d)
-To: $username
-
-Test mail for smarthost credentials.
-EOF
-        exit $?
+        [[ $tls = 'true' ]] && tls='--tls' || tls='--tls-optional-strict'
+        /usr/bin/swaks -a -q AUTH -au $username -ap $password --server $hostname --port $port $tls
```
DavidePrincipi, line 51: Protect shell special chars.

```suggestion
        /usr/bin/swaks -a -q AUTH -au "$username" -ap "$password" --server "$hostname" --port "$port" "$tls"
```

DavidePrincipi: Does it work with passwords like

    pass"word
    pass ; rm /word
    pass & word
    ...

DavidePrincipi (approved):

PR #349, opened 2022-01-11
Fix the user password equal when passwordstrength is none

stephdl:
When the `passwordstrength` is set to `strong`, then we have a validation of password inside the UI of cockpit to put the `save/change` button to disabled to enabled
- respect strong format
- respect length > 7
- must be equal

however when the `passwordstrength` is set to `none` we have no validation, the` save/change` button is already allowed even if the password mismatch

we must respect the following to get the button save/change to disabled to enabled

- length >7
- password is equal in the two fields 

This is what this PR does

https://github.com/NethServer/dev/issues/6621

### DIFF
--- a/ui/src/components/system/Settings.vue
+++ b/ui/src/components/system/Settings.vue
@@ -119,7 +119,7 @@
           </label>
           <div class="col-sm-5">
             <button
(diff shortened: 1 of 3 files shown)


edospadoni (approved):

stephdl:
obsoleted by https://github.com/NethServer/nethserver-cockpit/pull/350

